PR #7, opened 2014-12-16
xaf auditfile export module for Odoo V8

diellenne:
https://github.com/OCA/l10n-netherlands/wiki/1.-Community-modules#l10n_nl_xaf_auditfile_export


StefanRijnhart:
In the first place, thanks for attending to this module! However, this module depends on xml_template. Is this module in an OCA repository? Otherwise, it will be a problem because the OCA does not allow adoption of modules which depend on non-OCA (and non-core) modules.

This is based on the Neobis module from lp:neobis-addons/6.1. Did you discuss the adoption of their module in the OCA with them? Do you have bad experiences contributing to their modules on Launchpad?

In any case, you need to retain the Neobis copyrights in the file headers in addition to your own new copyright claim.

diellenne:
We developed this module after discussing with Neobis that is not maintaining it anymore. For this reason we decided to develop a new version of it starting from their work as mentioned in the module's description. At that time the l10n-netherlands project was not started but to be totally sure I will contact them ASAP to speak about it.

For the module xml_template module I contacted C2C to ask about it (the module seems to be present just in their R&D addons on launchpad) but I had no reply yet. In between to make the module working I added it in a special section of onesteinbv GitHub (https://github.com/onesteinbv/odoo_melting_pot). If the OCA modules cannot be depending from external partner modules we have to find a solution.

StefanRijnhart:
xml_template: good strategy. If it's Camp2camp, you may as well already propose it to an appropriate repository (https://github.com/OCA/reporting-engine?). They can comment on it there but I don't think they'll mind.
Let us know what Neobis says, but in any case let them keep their copyrights.

l10n_nl_xaf_auditfile_export/LICENSE
```diff
@@ -0,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
```
StefanRijnhart: Maybe put this file in the root of the repository instead of in every module?

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+import sys
+from openerp import release
+import datetime
+import time
+import logging
+from openerp.modules import get_module_resource
+
+
+class auditfile_generate(osv.osv):
+    """
```
StefanRijnhart, line 39: osv.osv is deprecated since 7.0. Please use orm.Model

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+        _file = get_module_resource(
+            'l10n_nl_xaf_auditfile_export', 'data', 'auditfile_template.xml')
+        if not _file:
+            raise osv.except_osv(_('Data Error'), _('Invalid build-in template'))
+
+        return open(_file).read()
+
+    _defaults = {
+        'template_content': _default_template,
+        'name': 'auditfile %s' % datetime.datetime.now().date(),
```
StefanRijnhart: Make this aware of the user's timezone using fields.date.context_today

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    _constraints = [(_check_dates, _('Please, check periods'),
+                    ['period_start', 'period_stop'])]
+
+    # creates auditfile
+    def create_auditfile(self, cr, uid, ids, id):
+        self._logger.info('create auditfile')
+
+        _date_format = '%Y-%m-%d'
```
StefanRijnhart: Instead of hardcoding, import DEFAULT_SERVER_DATE_FORMAT from openerp.tools

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+    """
+
+    _logger = logging.getLogger('auditfile_generate')
+
+    _name = "auditfile.generate"
+    _description = "Auditfile_export generation"
+
+    def _check_state(self, cr, uid, ids, state, args, context={}):
+        res = {}
+        for id in ids:
```
StefanRijnhart: Don't use reserved word 'id', but something like 'res_id' instead.

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+        'name': 'auditfile %s' % datetime.datetime.now().date(),
+        'period_start': 1,
+        'period_stop': 12,
+    }
+
+    _constraints = [(_check_dates, _('Please, check periods'),
+                    ['period_start', 'period_stop'])]
+
+    # creates auditfile
+    def create_auditfile(self, cr, uid, ids, id):
```
StefanRijnhart: 'id' argument should be context=None, by the looks of it.

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+
+        data = self.browse(cr, uid, ids[0], context=id)
+
+        # partners search
+        partners_obj = self.pool.get("res.partner")
+        partners_obj_ids = partners_obj.search(
+            cr, uid, ['|', ('supplier', '=', 'True'), ('customer', '=', 'True')]
+        )
+        partners = partners_obj.browse(cr, uid, partners_obj_ids)
+
```
StefanRijnhart, line 117: Pass context

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+                    _('Data Error'), _('Following companies have no country code: %s')
+                    % (', '.join(names))
+                )
+            return True
+        _validate_country_ids(partners)
+
+        # periods search
+        period_obj = self.pool.get("account.period")
+        period_ids = period_obj.search(
+            cr, uid, ['&', ('id', '>=', data.period_start.id), ('id', '<=', data.period_stop.id)]
```
StefanRijnhart: A search by dates seems more appropriate.

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+        )
+        periods = period_obj.browse(cr, uid, period_ids)
+
+        # journal search
+        journal_obj = self.pool.get("account.journal")
+        journal_ids = journal_obj.search(cr, uid, [])
```
StefanRijnhart: again, pass context here and everywhere below.

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+        journals = journal_obj.browse(cr, uid, journal_ids)
+
+        # accounts search
+        account_obj = self.pool.get("account.account")
+        account_ids = account_obj.search(cr, uid, [])
+        accounts = account_obj.browse(cr, uid, account_ids)
+        coa_ids = account_obj.search(cr, uid, ['&', ('user_type', '=', 1), ('parent_id', '=', False)])
```
StefanRijnhart: Replace hardcoded account_account_type id by xml id or something else that is more reliable.

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+
+            # attaches xml to module
+            template.attach_xml(
+                cr, uid, template_id, attach_to=data, xml=xml_new, name=data.name + _output_file,
+                fname=data.name + _output_file,
+                description='XML file representing this partner',
+            )
+            self._logger.info('auditfile_export created succesfully')
+            self.write(cr, uid, ids[0],
+                       {'state': 'done', 'date_printed': time.strftime('%Y-%m-%d %H:%M:%S')}, context=id)
```
StefanRijnhart: Void write to a function field 'state' without an inverse function.
Date format: import DEFAULT_SERVER_DATETIME_FORMAT from openerp.tools.

l10n_nl_xaf_auditfile_export/auditfile_export.py
```diff
@@ -0,0 +1,290 @@
...
+
+        except Exception, e:
+            self._logger.error(sys.exc_info())
+            raise osv.except_osv(_('Data Error'), _('Invalid Data.\n%s\nPlease, see logs for details.') % e)
+
+        self._logger.info('unlink xml_template')
+        template.unlink(cr, uid, template_id)
+        return True
+
+auditfile_generate()
```
StefanRijnhart: You don't need to instantiate ORM objects since 6.1 I believe.

l10n_nl_xaf_auditfile_export/auditfile_export.xml
```diff
@@ -0,0 +1,86 @@
...
+#
+##############################################################################
+ -->
+<openerp>
+    <data>
+
+        <record model="ir.ui.view" id="auditfile_tree_view">
+            <field name="name">auditfile.tree</field>
+            <field name="model">auditfile.generate</field>
+            <field name="type">tree</field>
```
StefanRijnhart: 'type' field is derived from the content of arch since 7.0. Specifying it explicitely is therefore deprecated and triggers warnings.

l10n_nl_xaf_auditfile_export/auditfile_export.xml
```diff
@@ -0,0 +1,86 @@
...
+                    <field name="state"/>
+                    <field name="date_printed"/>
+                </tree>
+            </field>
+        </record>
+
+        <record model="ir.ui.view" id="auditfile_form_view">
+            <field name="name">auditfile.form</field>
+            <field name="model">auditfile.generate</field>
+            <field name="type">form</field>
```
StefanRijnhart: Remove 'type' field here as well.

l10n_nl_xaf_auditfile_export/i18n/nl.po
```diff
@@ -0,0 +1,186 @@
...
+
+#. module: xaf_auditfile_export
+#: code:addons/xaf_auditfile_export/auditfile_export.py:287
+#, python-format
+msgid ""
+"Invalid Data.\n"
+"%s\n"
+"Please, see logs for details."
+msgstr ""
+"Data ongeldit.\n"
```
StefanRijnhart: Typo

diellenne:
Hi Stefan,

thank you for your help and for the effort you are spending in the review of this modules.
I will perform the fixes during the next week once I will come back to work after the Christmas holidays.

In between the review of this module is assigned to me (at the moment can just be in this way because I am the only member of the team); if you want to be directly involved in the merging/approval process you should be explicitly included in the maintainers team. There is a dedicated post in the issues section.

Cheers.

diellenne:
Neobis agrees with the inclusion of the module in the project. They are already mentioned in the module's description but I will also add them in all the copyright headers.

About the module xml_template I contacted C2C to ask about it...I'm waiting for a reply and then I will decide how to handle the problem.

StefanRijnhart:
Great, thanks for the follow-up. When looking through the code in xml_template, it seems like it is not a module of great quality. Maybe there is a better solution to create the xml file? Maybe you could look at how Alexis de Lattre generated SEPA XML files in https://github.com/OCA/bank-payment/tree/8.0/account_banking_sepa_credit_transfer.

diellenne:
I'm agree with you and I'm going to take a look at it ASAP but at the moment I think we can include the module as is (after the end of the review) and improve it in the future. No replies from C2C  about 'xml_template' so I'm going to submit their module to another OCA pertinent project as you suggested.

StefanRijnhart:
What I mean is, you will need to clean it up before it qualifies for inclusion and that may be more work than change the way of generating the XML.

pankk:
I reviewed and applied the changes suggested by Stefan in 3010f473e23b2c7b36e1631c33aeca6db154d605 with the exception of the "# periods search" part where Stefan suggested a search by dates. The audit export is based on periods, but maybe I missed something. Could you elaborate on why a search by dates would be more appropriate, Stafan?

StefanRijnhart:
Thanks @pankk, looking good!

Period search: yes, the report is by period. But in retrieving the periods between the start and end period, you rely on the id column of the period table to be in the same order as the start dates and end dates of the periods. I would not search by id, but by the start and end dates of the periods.

Something like the following, in https://github.com/OCA/l10n-netherlands/pull/7/files#diff-3e32dddeeef7815489748b59d90eda83R140:

```
['&', ('date_start', '>=', data.period_start.date_start), ('date_stop', '<=', data.period_stop.date_stop)]
```

diellenne:
Replaced by new module

PR #14, opened 2015-02-17
[ADD] l10n_nl_xaf_auditfile_export

hbrunn:
This is a proof of concept to use the qweb rendering engine for creating the auditfile. This should be faster and more robust than relying on `xml_template`.

While being on it, some features:
- multicompany support
- support _huge_ amounts of customers/suppliers/mutations


hbrunn:
My favorite side effect of this is that other modules can alter audit file generation by simply overriding the template

hbrunn:
This is feature complete now

diellenne:
Hi @hbrunn,
thank you for your effort in proposing an alternative approach for the xaf export of the financial info. From what I can see this is a new modules that you implemented from scratch and has to be tested, reviewed and verified to check if the generated file is totally compliant with the schema provided by the tax office. When @StefanRijnhart informed me about what you wanted to do I was thinking about modifications on the one already proposed to be merged in PR #7: for this kind of things is necessary to discuss it before start working and update the roadmap properly to not waste time and energy.

hbrunn:
well, just as #7 has to be tested, reviewed and verified. And one of the issues the review revealed was that the dependency on xml_template is not acceptable as it's not an OCA module. So here we are. Better test, review and verify this code.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -0,0 +1,253 @@
...
+            'and (company_id=%s or company_id is null)',
+            (tuple(p.id for p in self.get_periods()), self.company_id.id))
+        return self.env.cr.fetchall()[0][0]
+
+    @api.multi
+    def get_journals(self):
+        '''return journals'''
+        return self.env['account.journal'].search([
+            '|',
+            ('company_id', '=', False),
```
StefanRijnhart: company_id is required on journals, so you don't have to search for these.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -0,0 +1,253 @@
...
+        offset = 0
+        period_ids = [p.id for p in self.get_periods()]
+        while True:
+            results = self.env['account.move'].search(
+                [
+                    ('period_id', 'in', period_ids),
+                    ('journal_id', '=', journal.id),
+                    '|',
+                    ('company_id', '=', False),
+                    ('company_id', '=', self.company_id.id),
```
StefanRijnhart: company_id on account.move is a related field on the journal. Do you need this filter?

diellenne:
The #7 is already tested (technically and functionally), reviewed and adjusted: the only reason why is not merged yet is the dependency with the xml_template module and we was discussing how to go over it (thinking about a coding spring with Stefan). With that I don't want absolutely say that is better: I'm sure that your module is based on the native QWeb format is indeed more robust and modern  (than the other one that is based on an old one), but has to be also verified on the functional point of view to be sure that the generated file will be correct for the tax office. If he is Agree @StefanRijnhart can make a first technical review of it and after that somebody else (functional) can make a check on the generate file.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -0,0 +1,253 @@
...
+            ('company_id', '=', False),
+            ('company_id', '=', self.company_id.id),
+        ])
+
+    @api.multi
+    def get_taxes(self):
+        '''return taxes'''
+        return self.env['account.tax'].search([
+            '|',
+            ('company_id', '=', False),
```
StefanRijnhart: company_id is required

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -0,0 +1,253 @@
...
+        ])
+
+    @api.multi
+    def get_periods(self):
+        '''return periods in this export'''
+        return self.env['account.period'].search([
+            ('date_start', '<=', self.period_end.date_stop),
+            ('date_stop', '>=', self.period_start.date_start),
+            '|',
+            ('company_id', '=', False),
```
StefanRijnhart: company_id is required. Maybe you want to exclude speclal = True (opening periods)?

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -0,0 +1,253 @@
...
+                yield result
+            results.env.invalidate_all()
+            del results
+
+    @api.multi
+    def get_accounts(self):
+        '''return browse record list of accounts'''
+        return self.env['account.account'].search([
+            '|',
+            ('company_id', '=', False),
```
StefanRijnhart: company_id = required.

l10n_nl_xaf_auditfile_export/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Configuration
+=============
+
+This module works on huge amount of data, so there is a possibility to encounter out of memory exceptions. In this case. set the config parameter `l10n_nl_xaf_auditfile_export.max_records` to a value much lower than 10000.
+
+Known issues / Roadmap
+======================
+
+* encrypted and compressed files would be nice
```
StefanRijnhart, line 30: Looks like you don't cover a migration for previous installations, even though this seems necessary because the model and some field names changed. Can you mention this at least as a known issue?

StefanRijnhart:
Thanks @hbrunn. It is a bit sad for all people who worked on the previous incarnation of this module (including me, and I was a bit surprised to see this too), but I can see that the result of this reimplementation is of a higher level overall.

I have some comments inline, and can we have a pot file please?

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
hbrunn: I'm not so sure about that. We only export transactions, so this could mess up the balance I fear. The special periods are the ones where we make correction bookings, right?

l10n_nl_xaf_auditfile_export/README.rst
hbrunn, line 30: I did nothing here because there's 7.0 in between. And depending on which implementation we choose there (either backport this which would have my preference or port nbs_xml_auditfile_financieel) we need a different migration

hbrunn:
pot file added

diellenne:
If the module is now technically ok, during this week I will test it - also with an accountant - to check if it is correct from the functional point of view and if the generated file is compliant with the schema provided by the Dutch government. Cheers.

HiJohan:
Test the export xaf file. Important is the testdatabase. Most problem appear by Missing data. After this was adjusted we can generate the XAF file and export. 
Also tried to import this file in a XAF viewer without any problem.
Best regards

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
StefanRijnhart: I guess it depends on what you want to export but if it is about the balance, then I think you need to start from an opening journal, so excluding them entirely is indeed not an option.

l10n_nl_xaf_auditfile_export/README.rst
StefanRijnhart, line 30: OK, agreed.

StefanRijnhart:
Thanks @hbrunn for the followup, and @HiJohan for the functional validation. :+1: then.

diellenne:
Thank you @HiJohan for the validation! Also for me it seems to be ok! :+1:

diellenne:
@StefanRijnhart If for you the module is ok I'm going to make a final test and merge it during this week. Cheers.

StefanRijnhart:
Yes please!

diellenne:
:+1:

PR #19, opened 2015-08-18
8.0 l10n_nl_postcodeapi

StefanRijnhart:


gurneyalex:
pypostcode installed on runbot

StefanRijnhart:
Thank you @gurneyalex!

gurneyalex:
nice icon :smile_cat: 

:+1:

l10n_nl_postcodeapi/models/res_partner.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @ormcache(skiparg=2)
+    def get_provider_obj(self):
+        self._cr.execute("""
+                SELECT value FROM ir_config_parameter
+                WHERE key = %s""", ('l10n_nl_postcodeapi.apikey',))
```
hbrunn: any readson not to use `self.env['ir.config_parameter'].get_param`?

l10n_nl_postcodeapi/models/res_partner.py
```diff
@@ -0,0 +1,111 @@
...
+from lxml import etree
+from openerp import models, api, exceptions, _
+from openerp.tools import ormcache
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @ormcache(skiparg=2)
+    def get_provider_obj(self):
```
hbrunn, line 34: decorate with `@api.model`

l10n_nl_postcodeapi/models/res_partner.py
```diff
@@ -0,0 +1,111 @@
...
+        self._cr.execute("""
+                SELECT value FROM ir_config_parameter
+                WHERE key = %s""", ('l10n_nl_postcodeapi.apikey',))
+        row = self._cr.fetchone()
+        if row and row[0] != 'Your API key' and row[0].strip():
+            from pyPostcode import Api
+            provider = Api(row[0].strip())
+            test = provider.getaddress('1053NJ', '334T')
+            if not test or not test._data:
```
hbrunn: why that?

l10n_nl_postcodeapi/models/res_partner.py
```diff
@@ -0,0 +1,111 @@
...
+                    _('Error'),
+                    _('Could not verify the connection with the address '
+                      'lookup service (if you want to get rid of this '
+                      'message, please rename or delete the system parameter '
+                      '\'l10n_nl_postcodeapi.apikey\').'))
+            return provider
+        return False
+
+    @ormcache(skiparg=1)
+    def get_province(self, province):
```
hbrunn, line 53: `@api.model` and why skipping only one argument?

l10n_nl_postcodeapi/models/res_partner.py
```diff
@@ -0,0 +1,111 @@
...
+        keyword in Python
+        """
+        postal_code = self.zip and self.zip.replace(' ', '')
+        if (not (postal_code and self.street_number)
+                or (self.country_id and
+                    self.country_id != self.env.ref('base.nl'))):
+            return {}
+
+        provider_obj = self.get_provider_obj()
+        if not provider_obj:
```
hbrunn, line 76: I think the warning from above belongs here

Edit: No, in line 80

l10n_nl_postcodeapi/models/res_partner.py
```diff
@@ -0,0 +1,111 @@
...
+        not in a view."""
+        res = super(ResPartner, self).fields_view_get(
+            view_id=view_id, view_type=view_type,
+            toolbar=toolbar, submenu=submenu)
+        if view_type != 'form':
+            return res
+
+        def inject_onchange(arch):
+            arch = etree.fromstring(arch)
+            for field in ['zip', 'street_number']:
```
hbrunn: I'm surprised you need this in the first place (I'm curious: Why isn't the onchange decorator sufficient?), but if you need it, why don't you need it for country_id?

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart, line 34: That is just for backwards compatibility with old api. Do we need that?

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart: Invalid API key.

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart, line 53: Do we need backwards compatibility, and if I skip the province then nothing gets cached if I am correct.

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart, line 76: Then you get an exception every time that you enter an address, instead of when you change the API key.

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart: Good point.

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart: Because https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/res/res_partner.py#L68 does a hard replace of these fields, dropping any on_change.

StefanRijnhart: And modules dealing with street_number are likely to do the same.

StefanRijnhart: But you could be right, it might be wise to add country_id here.

l10n_nl_postcodeapi/models/res_partner.py
hbrunn, line 34: From somewhere I have in my head that non-decorated function will be `@api.multi`, but I can't look that up now. In general, I think it's good style to always use decorators so that we don't have to gues what `self` is. If you disagree, skip.

l10n_nl_postcodeapi/models/res_partner.py
hbrunn: then it errors out later. But this way, every `on_change` triggers two requests, that's not desireable, is it?

l10n_nl_postcodeapi/models/res_partner.py
hbrunn, line 53: very unscientific experiments once let me think `ormcache` counts old api parameters. https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L823 seems to support this voodoo-it.

what is definitely the case is that `self` doesn't count

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart: That is why the provider object is cached. This is only called once, after a restart of the server.

l10n_nl_postcodeapi/models/res_partner.py
hbrunn: ah, thanks

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart, line 53: I see, thanks. That needs fixing.

l10n_nl_postcodeapi/models/res_partner.py
hbrunn, line 76: now I get it, forgot about the caching. ignore the guy talking nonesense. But then as your test addres better use binnenhof 1a in den haag or the like, because the address you use is somewhat volatile to my knowledge

l10n_nl_postcodeapi/models/ir_config_parameter.py
```diff
@@ -0,0 +1,52 @@
...
+#
+##############################################################################
+
+from openerp import models, api
+
+
+class IrConfigParameter(models.Model):
+    _inherit = 'ir.config_parameter'
+
+    @api.model
```
hbrunn, line 30: this needs a returns decorator + downgrade, otherwise leagacy code calling your function will fail miserably: https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L4040

l10n_nl_postcodeapi/models/res_country_state.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        Clear the postcode provider cache when the state
+        table is altered.
+        """
+        self.env['res.partner'].get_province.clear_cache(
+            self.env['res.partner'])
+        return super(ResCountryState, self).write(vals)
+
+    @api.model
+    def create(self, vals):
```
hbrunn, line 40: @api.returns

StefanRijnhart:
Thanks @hbrunn for the thorough review! I think I fixed all the remaining issues.

l10n_nl_postcodeapi/models/res_partner.py
StefanRijnhart, line 76: FYI, the call will succeed as long as there is a zip code 1053NJ with house number 334. The suffix is not taken into account by the provider. So the test zip code should be safe.

l10n_nl_postcodeapi/models/ir_config_parameter.py
StefanRijnhart, line 30: While github does not seem to think the diff is outdated, there is a followup change to this comment in the line below.

hbrunn:
:+1:

PR #31, opened 2016-01-04
[FIX] show empty string if lastname is unset

hbrunn:
without this, we get 'False' as default name on a form where none of the fields involved is shown


StefanRijnhart:
:+1:

RoelAdriaans:
:+1:

elicoidal:
:+1: Fast tracked for obvious.

PR #35, opened 2016-02-15
[FIX] don't use private fields, use the API

hbrunn:
version 0.2 of the postcode api supports as well protocol version 1 as 2, and in version 2, the internal data structures are different. By sticking to what the API offers, the addons works in both cases.


StefanRijnhart:
:+1:

RoelAdriaans:
:+1:

PR #36, opened 2016-02-18
[FIX] Use right encoding when generating XML File

RoelAdriaans:
File should have UTF-8 as encoding, not utf8.

We had errors with reading the generated XAF file with the old encoding, solved it by changing the encoding to UTF-8. This is used in python too.


StefanRijnhart:
:+1: apparently, the XML specs say any application **should** support a lower case encoding (not: **must**) so this seems appropriate

hbrunn:
:+1:

hbrunn:
@RoelAdriaans-B-informed could you rebase on 8.0? then travis should become green

PR #37, opened 2016-02-27
Fix for travis test_flake8

astirpe:
This fix is to avoid the following test failure on Travis:

```
4.27s$ travis_run_tests
======== Testing test_flake8 ========
l10n_nl_postcodeapi/models/res_partner.py:71:17: W503 line break before binary operator
```

### DIFF
--- a/l10n_nl_postcodeapi/models/res_partner.py
+++ b/l10n_nl_postcodeapi/models/res_partner.py
@@ -67,9 +67,9 @@ class ResPartner(models.Model):
         keyword in Python
         """
         postal_code = self.zip and self.zip.replace(' ', '')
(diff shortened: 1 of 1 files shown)


hbrunn:
I think the whole expression would be a lot more readable if you removed unnecessary brackets and replaced the outermost one by a backslash:

``` python
        if not (postal_code and self.street_number) or\
                self.country_id and self.country_id != self.env.ref('base.nl'):
```

hbrunn:
:+1:

RoelAdriaans:
:+1:

nielsmiddeldorp:
:+1:

PR #38, opened 2016-03-01
[ADD] module l10n_nl_bsn 8.0

astirpe:
This module adds the BSN (Burgerservicenummer) field on partner forms.

I propose to use this module as a common basis for further developments.
Any feedback would be very appreciated.

### DIFF
--- a/l10n_nl_bsn/README.rst
+++ b/l10n_nl_bsn/README.rst
@@ -38,9 +38,6 @@ To use this module, you need to:
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/176/8.0
 
-.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
-.. branch is "8.0" for example
-
 
 Bug Tracker
 ===========
--- a/l10n_nl_bsn/README.rst
+++ b/l10n_nl_bsn/README.rst
@@ -1,0 +1,83 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+======================================
+Burgerservicenummer (BSN) for Partners
+======================================
+
+This module adds the BSN (Burgerservicenummer) field on partner forms.
+
+The field is visible when the flag is_company is false.
+
+A double check on the BSN is done when inserting/modifying its value:
+ - validation of the BSN (check whether the format is correct);
+ - check if another partner with the same BSN already exists.
+In both cases, a non-blocking alert is shown.
+
+
+Installation
+============
+
+The module depends on the external library 'python-stdnum'.
+
+You can install that library by using pip:
+
+* pip install python-stdnum
+
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Open a form of a contact, eg.: a person (uncheck the flag "Is a Company?")
+* Enter a BSN number
+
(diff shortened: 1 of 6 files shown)


l10n_nl_bsn/models/res_partner.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+# © 2016 ONESTEiN BV (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models, api
+from openerp.tools.translate import _
+from stdnum.nl import bsn
```
RoelAdriaans: Please use an import that does not fail when the module is not available.
See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md , External dependencies

l10n_nl_bsn/README.rst
```diff
@@ -0,0 +1,83 @@
...
+To use this module, you need to:
+
+* Open a form of a contact, eg.: a person (uncheck the flag "Is a Company?")
+* Enter a BSN number
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/176/8.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
```
RoelAdriaans: This notice can be removed

hbrunn:
:+1:

RoelAdriaans:
:+1:

lfreeke:
:+1: (test)

StefanRijnhart:
I know this has already been merged (too soon, for my taste), but there are legal aspects that need to be taken into account. It is forbidden to use the BSN if this is not required by a legal context. See https://autoriteitpersoonsgegevens.nl/nl/onderwerpen/identificatie/burgerservicenummer-bsn. One such legal context is HR management, so this change would make much more sense on the employee model instead of the partner model. Can you at least mention the legal considerations in the README?

hbrunn:
@StefanRijnhart good point, I did not think about this aspect at all, shame on me

astirpe:
Hi, as per the link above, the BSN can be user by some kind of organizations:

> Dit geldt bijvoorbeeld voor zorgverleners, zoals huisartsen, apotheken en zorgverzekeraars

If we want to make Odoo more useful for these type of organizations, this module is a good addition in my opinion.

> It is forbidden to use the BSN if this is not required by a legal context.

Good point :+1: 
That indeed should be mentioned in the README!

hbrunn:
@astirpe will you PR a change to the readme? Of course, we need to keep this module, only people should be aware of the legal issues

astirpe:
Please check #39 
Thank you!

andreschenkels:
I would also expect the BSN to be available and usefull on the employee. For companies the BSN == VAT so when do we need this field on partners?

nielsmiddeldorp:
I would suggest that the BSN only shows on the partner if it not a company. BSN is not only for employees but also for customers that are a person. E.g. a client of a health Insurance company.

StefanRijnhart:
@nielsmiddeldorp valid use case, indeed.

astirpe:
I confirm that the BSN only shows on the partner if it not a company.

ehahouimohssine:
For security reasons this should be only visible to HR related roles. Otherwise this will be in violation to the security framework of WBP regarding the protection of persons info.

astirpe:
Thanks @ehahouimohssine !
Please check PR #40

PR #40, opened 2016-03-03
Make BSN visible to HR Officer users.

astirpe:
For security reasons this should be only visible to HR related roles. Otherwise this will be in violation to the security framework of WBP regarding the protection of persons info.

### DIFF
--- a/l10n_nl_bsn/README.rst
+++ b/l10n_nl_bsn/README.rst
@@ -26,6 +26,18 @@ You can install that library by using pip:
 * pip install python-stdnum
 
 
+Configuration
+=============
+
+For security reasons the BSN number should be only visible to HR related roles.
+Otherwise this will be in violation to the security framework of WBP regarding
+the protection of persons info.
+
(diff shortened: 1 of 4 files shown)


pankk:
One miniscule bug: When having a valid BSN number for a partner, then changing it to an invalid number, save and get the correct warning, finally change it back to the original valid number, you will get a "Another person (Administrator) has the same BSN [...]" warning.

astirpe:
Yess it's a bug, I will fix it asap! Or feel free to open a PR. Thank you @pankk !

hbrunn:
This is not a security measure but only UI. Either actually suppress reading the field as non-privileged user by overriding `read` or setting the group on the field too, or reword the readme

astirpe:
Separate issue #41 opened, to track the bug noticed by @pankk .
Thanks for all the reviews!

hbrunn:
:+1:

ehahouimohssine:
:+1:

Issue #41, opened 2016-03-09
l10n_nl_bsn, improve check: BSN unique


astirpe:
One minor bug: When having a valid BSN number for a partner, then changing it to an invalid number, save and get the correct warning, finally change it back to the original valid number, you will get a "Another person (Administrator) has the same BSN [...]" warning.

PR #42, opened 2016-03-10
[ADD] Porting l10n_nl_bsn to V9

astirpe:


hbrunn:
:+1:

max3903:
:+1:

RoelAdriaans:
Does this port have the same issue as #41? 
If so, maybe wait before merging until that error is fixed.

astirpe:
Yes, the same issue #41 occurs on this port.
It's a very minor issue. It would probably never occur to an end user because, to reproduce it, you have to:
1. write a correct BSN and save
2. write a wrong BSN without saving
3. write the same BSN as point 1.

Is it a blocking issue for you?

StefanRijnhart:
:+1:

PR #43, opened 2016-03-13
Improve search for another partner with same BSN

### LINKED ISSUE #41: l10n_nl_bsn, improve check: BSN unique

astirpe:
Fixes #41

### DIFF
--- a/l10n_nl_bsn/models/res_partner.py
+++ b/l10n_nl_bsn/models/res_partner.py
@@ -37,16 +37,18 @@ class ResPartner(models.Model):
                     }
 
                 # search for another partner with the same BSN
-                args = [('bsn_number', '=', partner.bsn_number)]
+                args = [('bsn_number', '=', partner.bsn_number),
+                        ('name', '!=', partner.name)]
                 # refine search in case of multicompany setting
                 if partner.company_id:
                     args += [('company_id', '=', partner.company_id.id)]
-                partners = self.search(args, limit=1)
+                other_partner = self.search(args, limit=1)
                 # is another partner exists, display a warning
-                if partners:
+                if other_partner:
                     msg = _('Another person (%s) has the same BSN (%s).')
                     warning = {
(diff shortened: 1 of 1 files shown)


l10n_nl_bsn/models/res_partner.py
```diff
@@ -37,16 +37,18 @@ def onchange_bsn_number(self):
...
                 # search for another partner with the same BSN
-                args = [('bsn_number', '=', partner.bsn_number)]
+                args = [('bsn_number', '=', partner.bsn_number),
+                        ('name', '!=', partner.name)]
                 # refine search in case of multicompany setting
                 if partner.company_id:
                     args += [('company_id', '=', partner.company_id.id)]
-                partners = self.search(args, limit=1)
+                other_partner = self.search(args, limit=1)
                 # is another partner exists, display a warning
```
pankk: Typo, 'is' should be 'if' :)

pankk:
:+1:  (code review only)

hbrunn:
can't you search for the partner's id in the case it's available?

astirpe:
@hbrunn I've chosen this solution for 2 reasons, one is technical and the other is functional.

**Technical**: searching in an onchange method, I cannot read my current record id, because _self.id_ is always _NewId_. This occurs also in case the record was already stored in the database.

**Functional**: let's say, for any reason, you have some duplicated partners (individuals with the same name) and you want to keep them as they are. In this case, there is no reason to raise a warning, saying that another person has the same BSN. In facts, the same duplicated partners should get the same BSN.

hbrunn:
@astirpe thanks, makes sense. :+1:

StefanRijnhart:
:+1:

PR #44, opened 2016-03-21
[IMP] Use API 2, because old api is no longer supported

RoelAdriaans:
Until pyPostcode is updated to v0.3 on the odoo server, it still uses the now defunct version 1 api as default.

Added a warning that the old and new api keys are not compatible.

See https://github.com/steffex/pyPostcode/issues/12 and https://github.com/steffex/pyPostcode/pull/10


hbrunn:
:+1:

StefanRijnhart:
:+1:

PR #47, opened 2016-06-10
Added module l10n_nl_coa: new NL chart of accounts

astirpe:
This module provides a more generic and complete chart of accounts, in conjunction with a overall review of the module. New taxes for service (diensten) are added in the tax template.
The compatibility with the actual NL Vat Statements report is kept.


NL66278:
I would request to split the module in two. One for  things like tax tags and fiscal positions that are the same, no matter what kind of chart of accounts you use.  This could be part of the l10n_nl module were originally proposed. Then this module can add the chart of accounts as proposed here, and others (I volunteer) could propose a module for the chart of accounts according to the reference scheme. The user could then either install the one, or the other module (or even both in a multiple company scenario with different accounting needs).

StefanRijnhart:
Agree with @NL66278 that the improvements to the tax configuration can be submitted in https://github.com/odoo/odoo/issues/12351. For a new set of accounts, I remain that RSG is the way forward, as proposed now in https://github.com/OCA/l10n-netherlands/pull/48

astirpe:
This proposed module was meant to be a replacement of the standard localization module l10n_nl.

An initial PR (https://github.com/odoo/odoo/pull/12351) was opened towards Odoo V9.
Odoo SA accepted that proposed PR but decided to apply the replacement of module l10n_nl starting from Odoo V10.

So this actual PR could be considered as the back-porting to V9 of the functionality provided with l10n_nl (Odoo V10).

However, to avoid confusion and strange impacts that would occur in the future in case someone uses this module, I close this PR.
For every clarification, don't hesitate to contact me.

PR #48, opened 2016-06-16
[ENH] Add Reference Chart of Accounts template (RGS).

NL66278:
Give users the option to use the reference chart of accounts (Referentie Grootboek Schema).

See: http://www.referentiegrootboekschema.nl/home


lfreeke:
@NL66278 please check travis error

NL66278:
@lfreeke Travis errors solved. @StefanRijnhart  As you requested a module in the l10n-netherlands repo.

StefanRijnhart:
Thanks! Some initial feedback:

Can you fill the properties (such as property_account_receivable) which are defined in the regular Dutch chart of accounts?
Some accounts seem to have the wrong category. This is something I am not an expert in either, but I think 'Voorziening dubieuze debiteuren' is not a debtor account for instance, just like 'Vooruitbetalingen op bestelling' is not a creditor account. Maybe we should have the assignment of categories checked by an accountant.
In the account mapping in the Niet EU Landen fiscal position, there are double mappings. For instance, 8001010 maps to multiple accounts.

NL66278:
@StefanRijnhart Hello I corrected the debtor and creditor acounts.

There is no work to be done on properties. I think they are created using the information from the chart template (in account_chart_template.xml), this is no different from the original dutch chart of account.
I cannot find the double mappings in the fiscal position. And I do not use the names like 8001010 to refer to accounts, but the xml_id's. Where exactly did you see this?

hbrunn:
https://travis-ci.org/OCA/l10n-netherlands/jobs/138851702#L420

hbrunn:
:+1:

lfreeke:
:+1: (test) At first sight it all seems to be working.

PR #56, opened 2016-11-01
[ADD] l10n_nl_normalize_zip

hbrunn:
Normalize Dutch zip codes
=========================

This module was written to have zipcodes formatted in the standard Dutch way, which is 0000 XX.

Usage
=====

To use this module, you need to:

#. fill in a zip number on a partner
#. be happy that whatever you fill in there, if it looks remotely like a Dutch zip code, it will be forced to the format above.


hbrunn:
@pedrobaeza can you add server-tools to runbot?

astirpe:
@hbrunn with this module you format the zip for every partner regardless if it's country is Netherlands or not. Is it correct? In this case is it possible to format the zip depending on the Country if it's already set?

pedrobaeza:
server-tools added in runbot and rebuild launched

hbrunn:
@pedrobaeza thanks!
@astirpe I consider it correctish ;-) Note that the code is quite cautious about not touching strings that don't look like Dutch zip codes. A check for the first 4 being digits and the last two being characters wouldn't hurt I guess though, added this right now.

Judging from https://en.wikipedia.org/wiki/List_of_postal_codes, the Dutch format is unique, so it shouldn't clash with some other country and we avoid reading the partners for every write on the zip code.

NL66278:
:+1:  LGTM

andreschenkels:
:+1:

astirpe:
@hbrunn when installing the module, the zip value of existing partners is not automatically updated; was it intentional to not provide a post_init_hook for normalizing the zip for all the existing partners?

Code review is :+1:  for me

hbrunn:
intentional as in not relevant for my current problem. It's certainly a good thing to have for people who want to install this in an existing database, should be very straightforward to add

hbrunn:
can we merge this?

pankk (approved):
Reviewed code, looks great!

tarteo (approved):
👍

PR #57, opened 2016-12-02
[MIG] 9.0 l10n_nl_intrastat Dutch Intrastat report (ICP)

thomaspaulb:
Migrated and added tests.


l10n_nl_intrastat/README.rst
```diff
@@ -0,0 +1,76 @@
...
+Known issues / Roadmap
+======================
+
+* The Dutch tax authority accepts an XML format, generate this to avoid manual processing.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
hbrunn: replace `project_repo`

l10n_nl_intrastat/README.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Therp BV
```
hbrunn, line 71: how about yourself?

l10n_nl_intrastat/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+
+from . import models
+
```
hbrunn: empty line

l10n_nl_intrastat/__openerp__.py
```diff
@@ -0,0 +1,62 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
hbrunn: use short license headers: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py#L1-L3, but keep the dates and people involved

l10n_nl_intrastat/__openerp__.py
```diff
@@ -0,0 +1,62 @@
...
+
+{
+    'name': 'Intrastat reporting (ICP) for the Netherlands',
+    'version': '9.0.1.0.0',
+    'category': 'Localisation/Report Intrastat',
+    'license': 'AGPL-3',
+    'description': """
```
hbrunn: move this to `README.rst`, and add a key `summary` instead

l10n_nl_intrastat/i18n/l10n_nl_intrastat.pot
```diff
@@ -0,0 +1,172 @@
+# Translation of OpenERP Server.
```
hbrunn: we don't need pot files any more

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields, api, _
+from openerp.exceptions import Warning
```
hbrunn: adapt this to the 6th bullet in https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#idioms

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+        invoice_domain = [
+            ('type', 'in', ('out_invoice', 'out_refund')),
+            ('date_invoice', '>=', self.date_from),
+            ('date_invoice', '<=', self.date_to),
+            ('state', 'in', ('open', 'paid')),
+            ('company_id', '=', self.company_id.id),
+        ]
+        # Signal invoices without a country
+        # (Should not happen, as country_id on parter is set to required)
+        invalid_invoices = invoice_model.search_read(
```
hbrunn: what do you think about treating partners without country as domestic partners, inferred from the company's country?

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+                    inv['partner_id'][1] for inv in invalid_invoices)
+            )
+        # Search invoices that need intrastat reporting:
+        invoice_records = invoice_model.search(
+            invoice_domain + [
+                ('partner_id.country_id.intrastat', '=', True),
+                ('partner_id.country_id.id', '!=', company_obj.country_id.id),
+            ]
+        )
+        invoice_ids = [inv['id'] for inv in invoice_records]
```
hbrunn: use `invoice_records.ids`

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+        invoice_line_records = invoice_line_model.search(
+            [('invoice_id', 'in', invoice_ids)])
+
+        # Gather amounts from invoice lines
+        total_amount = 0.0
+        partner_amounts_map = {}
+        for line in invoice_line_records:
+            # Ignore invoiceline if taxes should not be included in intrastat:
+            if any(tax.exclude_from_intrastat_if_present for tax in
+                    line.invoice_line_tax_ids):
```
hbrunn: can't you add this to the search criteria above?

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+                continue
+            # Report is per commercial partner:
+            commercial_partner_id = (
+                line.invoice_id.partner_id.commercial_partner_id.id)
```
hbrunn: recordsets return their id list as hash, so there's no need to get the id: https://github.com/OCA/OCB/blob/9.0/openerp/models.py#L5736 (but then of course you need to be consequent and use the recordset everywhere)

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+            if commercial_partner_id not in partner_amounts_map:
+                partner_amounts_map[commercial_partner_id] = {
```
hbrunn: use `partner_amounts_map.setdefault`

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+                    'amount_product': 0.0,
+                    'amount_service': 0.0,
+                }
+            amounts = partner_amounts_map[commercial_partner_id]
```
hbrunn: and `setdefault` returns what you want here: https://docs.python.org/2/library/stdtypes.html#dict.setdefault

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+            # Determine product or service:
+            if (line.product_id and (
```
hbrunn: this is very 7.0 style programming. You don't need this check, because if this is not set, accessing the empty record set will yield falsy results

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+                    line.product_id.type == 'service' or
+                    line.product_id.is_accessory_cost)):
+                amount_type = 'amount_service'
+            else:
+                amount_type = 'amount_product'
+            sign = line.invoice_id.type == 'out_refund' and -1 or 1
+            amount = sign * line.price_subtotal
+            # Convert currency amount if necessary:
+            line_currency_obj = line.invoice_id.currency_id  # simplify
```
hbrunn: the `_obj` suffix is a bit misleading because most people use this to indicate this is a variable holding a model as opposed to a recordset. What's wrong with `currency`?

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+            invoice_date = line.invoice_id.date_invoice
+            if (line_currency_obj and
```
hbrunn: same as above. and recordsets have overridden equality to do what you want in the line below, no need to zoom into the ids: https://github.com/OCA/OCB/blob/9.0/openerp/models.py#L5692

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+                    line_currency_obj.with_context(date=invoice_date).compute(
+                        amount, company_obj.currency_id, round=True)
+                )
+            # Accumulate totals:
+            amounts[amount_type] += amount  # per partner and type
+            total_amount += amount  # grand total
+
+        # Determine new report lines
+        new_lines = []
+        for (partner_id, vals) in partner_amounts_map.items():
```
hbrunn: better use `.iteritems`, this function doesn't create a new copy of the dict

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,220 @@
...
+            if not (vals['amount_service'] or vals['amount_product']):
+                continue
+            vals.update({'partner_id': partner_id})
+            new_lines.append(vals)
+
+        # Remove existing lines
+        for line_obj in self.line_ids:
```
hbrunn: why not `self.line_ids.unlink()`?

hbrunn (changes requested):

l10n_nl_intrastat/models/l10n_nl_intrastat.py
thomaspaulb: It would make the module more friendly to implement in client installs, where there are typically only a few foreign partners and people don't want to always select country. Do you know why this wasn't done in the first place?

hbrunn: no idea why that's not done in the original, but let's do it here

l10n_nl_intrastat/models/l10n_nl_intrastat.py
thomaspaulb: Yep, have succeeded in that now.

l10n_nl_intrastat/models/l10n_nl_intrastat.py
thomaspaulb: Yes, but isn't it better to sacrifice concise coding for verbosity on this? The line is:

    if (line.product_id and (line.product_id.type == 'service' or line.product_id.is_accessory_cost))

If we are to change this to:

    if (line.product_id.type == 'service' or line.product_id.is_accessory_cost)

And someone decided to refactor this to 

    if (line.product_id.type == 'service' or not line.product_id.is_not_an_accessory_cost)

We will have introduced buggy behaviour for the case when there is no product.

hbrunn: that's up to the person doing the refactoring. With the same argument, you could want to name all default parameters of a function because somebody could change the defaults

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,193 @@
...
+        self._check_generate_lines()
+
+        # We need at least these variables to calculate
+        if not self.date_from or not self.date_to or not self.company_id:
+            return
+
+        # Define search for invoices for period and company:
+        company_obj = self.company_id  # simplify access
+        invoice_domain = [
+            ('type', 'in', ('out_invoice', 'out_refund')),
```
thomaspaulb: A question about this: I saw that the Odoo intrastat also includes in_invoice and in_refund. Should we do that too?

hbrunn: to my knowledge, intrastat only cares for what you sell (or take back), so filtering for `out_*` seems correct to me

l10n_nl_intrastat/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Intrastat reporting (ICP) for the Netherlands',
+    'version': '9.0.1.0.0',
+    'category': 'Localisation/Report Intrastat',
+    'license': 'AGPL-3',
+    'summary': """Intrastat report for the Netherlands""",
+    'author': 'Therp BV, Odoo Community Association (OCA)',
+    'website': 'https://launchpad.net/new-report-intrastat',
```
astirpe: Could we think about an updated website?
Maybe [https://github.com/OCA/intrastat](https://github.com/OCA/intrastat) ?

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,194 @@
...
+            "State of the declaration. When the state is set to 'Done', "
+            "the parameters become read-only."
+        ),
+    )
+    line_ids = fields.One2many(
+        string='ICP line',
+        comodel_name='l10n_nl.report.intrastat.line',
+        inverse_name='report_id',
+        readonly=True,
+        # compute='_compute_lines_and_total'
```
astirpe: Could you remove this commented line?

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,194 @@
...
+
+        # Check whether all configuration done to generate report
+        self._check_generate_lines()
+
+        # We need at least these variables to calculate
+        if not self.date_from or not self.date_to or not self.company_id:
+            return
+
+        # Define search for invoices for period and company:
+        company_obj = self.company_id  # simplify access
```
astirpe: Here if you replace the name of the variable from `company_obj` to `company` (and do the same for its usage in lines 98, 130, 132, etc..), then you are compliant with the [OCA guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#variable-names)

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,194 @@
...
+        invoice_domain = [
+            ('type', 'in', ('out_invoice', 'out_refund')),
+            ('date_invoice', '>=', self.date_from),
+            ('date_invoice', '<=', self.date_to),
+            ('state', 'in', ('open', 'paid')),
+            ('company_id', '=', self.company_id.id),
```
astirpe: Or:
`('company_id', '=', company_id.id),`
is also good, if you rename the variable from `company_obj `to `company` (see comment above).

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,194 @@
...
+            new_lines.append(vals)
+
+        # Set values and replace existing lines by new lines
+        self.line_ids.unlink()
+        self.write({
+            'last_updated': fields.Datetime.now(),
+            'line_ids': [(0, False, line) for line in new_lines],
+            'total_amount': total_amount
+        })
+
```
astirpe, line 167: It seems that method 
```
@api.multi
def unlink(self):
    """
    Do not allow unlinking of confirmed reports
    """
    ....
```
is not ported here to V9, but is present in the other PR for V8.
Is there any reason why was not ported?
I think is needed also here for V9.

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,194 @@
...
+    _inherit = ['intrastat.common']
+
+    last_updated = fields.Datetime(readonly=True, required=False)
+    date_range_id = fields.Many2one(
+        comodel_name='date.range',
+        required=True,
+        string='Date range'
+    )
+    date_from = fields.Date(related='date_range_id.date_start', readonly=True)
+    date_to = fields.Date(related='date_range_id.date_end', readonly=True,
```
astirpe: We have a `date_to` which is `stored=True`, instead `date_from` is not stored.
Could you check?

Please also evaluate to use not-related fields, instead of related fields, for `date_from` and `date_to`. Both fields should be updated with an onchange on field `date_range_id` or, alternatively, in create() and write() methods of the report model. This way both the date fields will be remain unmodified in case someone updates `date_range_id.date_start` and `date_range_id.date_end`.

l10n_nl_intrastat/tests/test_all.py
```diff
@@ -0,0 +1,195 @@
...
+            'name': 'Test date range type',
+            'company_id': company.id,
+            'allow_overlap': False
+        })
+
+        # Create a date range spanning the last three months
+        date_range = self.env['date.range'].create({
+            'name': 'FS2016',
+            'date_start': datetime.today() + relativedelta(months=-3),
+            'date_end': datetime.today(),
```
astirpe: Unit tests are failing in travis, could you check the [travis log](https://travis-ci.org/OCA/l10n-netherlands/jobs/207029684)?

I think the tests failure depends on these 2 lines: you are filling `date_start` and `date_end` with datetime objects. You should cast the datetime objects to `fields.Date` type instead, using `datetime.strftime()` for example. You already used `datetime.strftime()`, check the model `l10n_nl.report.intrastat` for the example.

astirpe (changes requested):
I've just made a code review.
I didn't test the module, I will test it when the unit tests will be running successfully.

@thomaspaulb , thanks for porting this module!

astirpe:
@thomaspaulb here my Company is interested in having this module ported to V10. Are you going to work on it the next days? Let me know if there is something I can do. Thank you!

thomaspaulb:
I think I have processed all the review items, I have indeed followed the suggestion to make the related date fields into normal ones, and to add the unlink protection again. I have no plans yet to work on the V10 version, but am happy to be of any assistance.

jvOrsouw:
I've tested this module in V10, but my invoice-amount didn't show up in the ICP report. In my test, the customer is from France, has fiscal position 'EU country', my company's country is 'the netherlands', and the company of the report is the same as the company of the invoice. And finally, the invoice date is in the date range of the ICP report. 
On the invoice, the tax code is automatically set to 'sales within EU', and the amount shows in the correct column in the Tax Statement.
When I empty the tax code from the invoice, it does show up on the ICP report. 

Please correct so that also amounts with tax codes 'sales within EU' (all sorts) are taken into the ICP report.

astirpe:
The issue mentioned by Julia seems fixed by just restoring the V8 code when defining the search domain on account.invoice.line model. See commit https://github.com/OCA/l10n-netherlands/pull/72/commits/1a917d0db66cb771bd502167982a25283e0c59f5

cc @thomaspaulb

jvOrsouw (approved):
Yes, all validated revenue is  now shown in this report. 

Great addition for dutch localization

👍

hbrunn:
@thomaspaulb could you fix https://travis-ci.org/OCA/l10n-netherlands/jobs/219660441#L482-L490? And @astirpe, do you approve meanwhile? The thing is that I'd like to ask you to follow https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0 for #72, and this only works conveniently if this is merged before, because it will probably involve squatting commits

astirpe:
Please give a look at my previous comment, to fix the issue found by Julia. Other than that, for me is ok to merge.

astirpe:
@thomaspaulb the version 10.0 of this module was merged last night, see #72.
That version contains some changes regarding the module name and description, it would be good if you can apply the same changes to within this PR (and perhaps to the PR for version 8.0).
Thank you!

thomaspaulb:
Hi, (on the plane) I backported most of the style and cosmetic improvements @astirpe made in #72 as well as the PDF report that was added. Most importantly though I added [this bugfix](https://github.com/OCA/l10n-netherlands/commit/1a917d0db66cb771bd502167982a25283e0c59f5) that was asked for. Tests are running and the module works good on my PC. Also when I run `diff -r` against the current 10.0 version of the module I see no important changes left.

astirpe (approved):
Hi @thomaspaulb, (From the airport of Barcelona) I've checked and reviewed the code. Thank you!

astirpe:
@thomaspaulb to fix the travis error, just add the following line:
[postgresql: "9.2" # minimal postgresql version for the daterange method](https://github.com/OCA/l10n-netherlands/blob/10.0/.travis.yml#L5)
to the `.travis.yml` file. This because the version used by travis is by default not compatible with the `date_range` module.

Anyway, it's something about travis settings and my approval for this PR still remain valid :+1:

thomaspaulb:
@hbrunn A lot was done since your change request :-) How does it look for you now

hbrunn (approved):

PR #58, opened 2016-12-02
[MIG] 8.0 l10n_nl_intrastat

thomaspaulb:
Imported from Launchpad, ported to v8, added tests.


hbrunn:
did you consider using this https://github.com/OCA/maintainer-tools/wiki/Migration-Launchpad-%E2%86%92-GitHub method to import from launchpad, this way keeping authors and commits?

thomaspaulb:
I saw it but I somehow thought it was only for migrating full repos, not single modules to add to an existing OCA repo. Is it useful to keep history for this one?

hbrunn:
Sometimes it's helpful for stuff, but it's also a lot of extra work. You'd have to filter the commits for the directory in question as outlined in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-8.0#how

Let's see what other reviewers think and do nothing for now

hbrunn:
@thomaspaulb can you rebase?

oca_dependencies.txt
```diff
@@ -1 +1,2 @@
 partner-contact https://github.com/OCA/partner-contact 8.0
+intrastat https://github.com/sunflowerit/intrastat 8.0-intrastat_base-tests
```
astirpe: Module `intrastat_base` is available in OCA for V8, so I think you can set the OCA repository `intrastat` here, instead of the sunflowerit branch. Or is it a temporary solution for some reason?

hbrunn: it was probably not merged at the time of the PR, so this was necessary to get the tests right. But by now, this should indeed just be 'intrastat'

hbrunn: (took the liberty to do exactly that in order to continue with this)

astirpe:
About keeping authors and commits when importing from launchpad, I think the decision is up to the Authors (@therp, @alexis-via, etc..).

hbrunn:
We don't really mind attribution wise as the license headers are kept intact.

For git migrations from older branches I'd enforce this, because it makes cherry picking stuff back and forth between the different versions a lot simpler. But this is not going to happen with launchpad, so technically, it wouldn't be useful.

@StefanRijnhart what do you think? It's your commits that are mutilated here

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,277 @@
...
+            'date_done': fields.Date.today(),
+            'state': 'done',
+        })
+
+    @api.multi
+    def unlink(self):
+        """
+        Do not allow unlinking of confirmed reports
+        """
+        if self.search([('state', '!=', 'draft')]):
```
astirpe, line 231: This porting of `unlink()` method would not work as expected: it actually checks "all" the records that are not in draft.
It should be instead something like:
```
for report in self:
    if report.state != 'draft':
        raise ...
```

l10n_nl_intrastat/__openerp__.py
```diff
@@ -0,0 +1,18 @@
...
+    'version': '8.0.1.0.1',
+    'category': 'Localisation/Report Intrastat',
+    'license': 'AGPL-3',
+    'summary': """Intrastat report for the Netherlands""",
+    'author': 'Therp BV, Odoo Community Association (OCA)',
+    'website': 'https://launchpad.net/new-report-intrastat',
+    'depends': ['intrastat_base'],
+    'data': [
+        'views/l10n_nl_intrastat.xml',
+        'security/ir.model.access.csv',
```
astirpe, line 16: security/* files are commonly declared before the views/* files

l10n_nl_intrastat/__openerp__.py
```diff
@@ -0,0 +1,18 @@
...
+# Modifications Copyright 2012-2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': 'Intrastat reporting (ICP) for the Netherlands',
+    'version': '8.0.1.0.1',
+    'category': 'Localisation/Report Intrastat',
+    'license': 'AGPL-3',
+    'summary': """Intrastat report for the Netherlands""",
+    'author': 'Therp BV, Odoo Community Association (OCA)',
+    'website': 'https://launchpad.net/new-report-intrastat',
```
astirpe, line 12: Could we think of an updated website to propose?
Maybe [https://github.com/OCA/intrastat](https://github.com/OCA/intrastat) ?

l10n_nl_intrastat/i18n/l10n_nl_intrastat.pot
```diff
@@ -0,0 +1,172 @@
...
+#. module: l10n_nl_intrastat
+#: model:ir.actions.act_window,name:l10n_nl_intrastat.l10n_nl_intrastat_tree_action
+#: model:ir.ui.menu,name:l10n_nl_intrastat.l10n_nl_intrastat_report_menu
+msgid "ICP Report"
+msgstr ""
+
+#. module: l10n_nl_intrastat
+#: field:l10n_nl.report.intrastat,date_done:0
+msgid "Date done"
+msgstr ""
```
astirpe, line 172: File .pot could be safely removed, is not required anymore

l10n_nl_intrastat/README.rst
```diff
@@ -0,0 +1,76 @@
...
+Known issues / Roadmap
+======================
+
+* The Dutch tax authority accepts an XML format, generate this to avoid manual processing.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
astirpe, line 46: The link here should be `https://github.com/OCA/l10n-netherlands/issues`

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,277 @@
+# -*- coding: utf-8 -*-
+"""Define intrastat report (ICP) for dutch tax authorities."""
```
astirpe, line 2: This comment line is already present in the comment of the class, so it can be just removed.

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,277 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
astirpe, line 26: It would be nice to have short header also for this file.

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def _default_company_id(self):
+        """Default company is active user company."""
+        return self.env.user.company_id.id
+
+    period_id = fields.Many2one(
+        string='Period',
+        comodel_name='account.period',
+        states={'done': [('readonly', True)]},
+        default=_default_period_id,
```
astirpe, line 61: Here:
`default=lambda self: self._default_period_id(),`

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,277 @@
...
+    )
+    company_id = fields.Many2one(
+        string='Company',
+        comodel_name='res.company',
+        required=True,
+        states={'done': [('readonly', True)]},
+        help="Related company.",
+        default=_default_company_id,
```
astirpe, line 69: Here:
`default=lambda self: self.env.user.company_id`

and method `_default_company_id()` can be removed.

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,277 @@
...
+        total_amount = 0.0
+        partner_amounts_map = {}
+        # Check whether all configuration done to generate report
+        self._check_generate_lines()
+
+        # Remove existing lines
+        for line_obj in self.line_ids:
+            line_obj.unlink()
+        # Define search for invoices for period and company:
+        company_obj = self.company_id  # simplify access
```
astirpe, line 140: Here if you replace the name of the variable from `company_obj` to `company` (and do the same for its usage in the subsequent lines), then you are compliant with the [OCA guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#variable-names)

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,277 @@
...
+        invoice_model = self.env['account.invoice']
+        invoice_line_model = self.env['account.invoice.line']
+
+        total_amount = 0.0
+        partner_amounts_map = {}
+        # Check whether all configuration done to generate report
+        self._check_generate_lines()
+
+        # Remove existing lines
+        for line_obj in self.line_ids:
```
astirpe, line 137: You can simply do this:
`self.line_ids.unlink()`

l10n_nl_intrastat/models/l10n_nl_intrastat.py
```diff
@@ -0,0 +1,277 @@
...
+        # Define search for invoices for period and company:
+        company_obj = self.company_id  # simplify access
+        invoice_domain = [
+            ('type', 'in', ('out_invoice', 'out_refund')),
+            ('period_id', '=', self.period_id.id),
+            ('state', 'in', ('open', 'paid')),
+            ('company_id', '=', company_obj.id),
+        ]
+        # Signal invoices without a country
+        # (Should not happen, as country_id on parter is set to required)
```
astirpe, line 148: Typo: `parter` -> `partner`
But real question is: why this comment line? `country_id` on partner is not required by default.
I think this comment line should be removed. I would prefer this way, instead of setting `country_id` as required on partner.

astirpe (changes requested):
@thomaspaulb thanks for porting this module.

Here is just a code review, I didn't test the module.

andreschenkels:
Do you guys need some help here?

thomaspaulb:
Hey @aschenkels-ictstudio, of course help is always appreciated :-) If I remember correctly, after the code sprint in Barcelona me and @astirpe managed to get the 9.0 and 10.0 versions accepted, but this one slipped my mind. @astirpe what did we suggest for this module.. backporting or resuming with this one?

hbrunn:
@thomaspaulb you already made https://github.com/OCA/l10n-netherlands/tree/9.0/l10n_nl_intrastat, don't you think it's simpler to backport that? Originally the 9 version cam from this I think to remember, so that might just work with cherry picking and changing some xmlids?

thomaspaulb:
@hbrunn I think so too. The 9.0 version was what we finished, reviewed and polished in the end, also the unit tests there are more developed. So it's best to just backport that, run the unit tests and do some fixing here and there. @aschenkels-ictstudio do I hear you volunteering for the job?

astirpe:
@thomaspaulb I think doing a backport from 9.0 is faster, but be careful there are some aspects to consider:

- The version 9.0 of this module makes use of `date_range` and gets rid of V8.0 fiscal periods. It means that, if you go for the backporting way, you will have to ensure that you continue to use periods instead of `date_range` (module `date_range` is not available in V8.0 and I don't think it will be ever backported).
- The Monetary field type used in V9.0 is not existing in V8.0

... and surely I'm missing something else. So at the end you will have to resume part of the code V8.0.

Whatever will be your choice, I will be happy to review it :)

hbrunn:
as of the discussion above I close this one

PR #59, opened 2017-02-23
[10.0][MIG] Port l10n_nl_xaf_auditfile_export

astirpe:
I'm interested in porting module `l10n_nl_xaf_auditfile_export` to V10.
This is a work in progress. In case someone already started to work on it and wants to open a different PR, I'm ready to close this one.

Module `l10n_nl_xaf_auditfile_export` for V8 exports accounting data referring to fiscal periods. But periods were removed from the accounting of Odoo since V9. As per the specifications of the XAF auditfile format, reference to fiscal periods seem required (for example `<periodNumber>`).

Does anyone already analyzed or found a solution about this issue?


hbrunn:
this looks good at first glance. Before diving into details, did you consider using https://github.com/OCA/server-tools/tree/10.0/date_range for the dates? Most accounting related OCA modules involving intervals use this, so I think it's more convenient for the user to use it here too

astirpe:
With this first commit, I assume that the module works with the following limitations:

- date range represents a fiscal year, by default the actual fiscal year
- periods are always monthly based
- period numbers (`<periodNumber>`) are coded with 3 digits: 1st digit is the last digit of the fiscal year, the other 2 digits are the month (Eg.: period "February 2017" is represented by number `702`)

astirpe:
@hbrunn Yes, `date_range` would be useful for easy setting initial and end dates of a fiscal year. Unfortunately doesn't really help for handling periods...

hbrunn:
How about extending date ranges with the fields you need? This way you don't need assumptions, and also support nonstandard accounting periods, because the user can define the ranges however she sees fit. Then this module (or some helper module) can define a fiscal period date range type, and then you're done. Let's call it `account_fiscal_period` and let it do the same thing as https://github.com/OCA/account-financial-tools/tree/10.0/account_fiscal_year, but then for periods

astirpe:
@hbrunn That idea was discussed during the Sorrento Code sprint for better generating fiscal reports; at the end we all agreed that reintroducing fiscal periods would be unwanted and undesirable.
On the other side, in some cases (like for this module), fiscal periods are needed.

In my opinion, if the limitations I just described above are well accepted from the NL Community, then we could think about keeping this module decoupled from an hypothetical `account_fiscal_period`. Instead if we really need a more flexible solution, then I'm ready to start or contribute to `account_fiscal_period`.

Thank you!

hbrunn:
Let's see what other Dutch community members say. I'd welcome such a module, as the concept of a financial period (in the sense of: I want to easily filter for what I consider my relevant periods for move lines) is hard to knock out of our customers' accountants' minds...

jvOrsouw:
I'm testing it. I went to accounting - settings and changed the last date of the fiscal year. Then when I opened the form for the auditfile, the end-date is still 31st of december. It should be the same as the end date in the accounting settings. The default start date should be 12 months before the end date.

- The periods in the generated auditfile have the wrong start and end date. Should be the start and end date of the month, and not of the fiscal year.

- Periods are always monthly based: I fully agree.

- Period number: that's a very good solution to create understandable period numbers especially with Odoo 10 not having period numbers, and the limitations of a period number maximum of 3 digits.

astirpe:
@jvOrsouw thank you for the feedback!

I've just tried in my local PC to change the last date of the fiscal year from the standard Odoo (accounting -> settings) and the module worked correctly. The default end date is the one set in the settings and the default start date is 12 months before the end date.
Anyway there is an OCA module `account_fiscal_year` that modifies the calculus of the dates, overriding the standard behavior, taking by default the date range defined for the fiscal year. May I ask you to check that `account_fiscal_year` is installed in your server and that the dates are set accordingly with your fiscal year configuration? You can find the date range settings in Settings -> Date Ranges -> Fiscal year 2017 (or similar). Please let me know!

About the second point, indeed the periods have the wrong start and end date! I'm working on it for a fix. Thank you!

And thanks for confirming the other 2 points!

jvOrsouw:
retested the issue: I noticed that when the module 'account_fiscal_year' is installed, the start and end date look at the fiscal year entered as 'date range'. When I deleted this range, the fiscal period changes to the fiscal year as entered in the accounting settings. That's exactly what I want, so that works exactly as it should.

- The bug with the start and end date of the periods is fixed, the dates are now the beginning and end date of the month.


this module is working well for me. 👍

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -55,39 +62,30 @@ def _auditfile_name_get(self):
...
     @api.one
-    @api.constrains('period_start', 'period_end')
-    def check_periods(self):
-        if self.period_start.date_start > self.period_end.date_start:
+    @api.constrains('date_start', 'date_end')
+    def check_dates(self):
+        if self.date_start > self.date_end:
             raise exceptions.ValidationError(
-                _('You need to choose consecutive periods!'))
+                _('Starting date must be anterior ending date!'))
```
espo-tony: According to the if condition, starting date can be also equal to ending date.
Please, either correct the condition or the error message.

espo-tony (commented):
Here's my review. Anyway I don't consider it blocking. :+1:

astirpe:
I've just moved the menuitem for the exporting action, because there was a crash when installing the module in Odoo Enterprise.

I've also enabled the feature for the groups 'Accountant' and 'Adviser', so that having just the 'Billing' rights is not enough. Please let me know if you agree...

CasVissers-360ERP:
I tested the module on an Odoo 10 Enterprise instance. the module works correctly. I'm not sure if adding the year to the period numbering is correct, but it's a nice solution for exporting multiple years. Time will tell.

PR #60, opened 2017-02-24
Add serbia to the list of countries

thomaspaulb:


hbrunn:
this looks like a no-brainer, but I'm a bit concerned that this might break at the receiving end. If the importing application does validation against the official xsd (http://www.softwarepakket.nl/swpakketten/auditfiles/auditfile_financieel.php?bronw=1 to my knowledge), files containing serbia won't work. Did you submit files containing this to the tax office and can they use it?

astirpe:
@hbrunn even if the tax office can use it, the point is that the spec file `XmlAuditfileFinancieel3.2.xsd` will be different than the "certified" one. This is not a technical issue, it's more a specification issue. I would like to help, but how?

hbrunn:
file a PR against the tax office's github repo? *ducks and runs*

seriously, I have no idea how to approach this on a technical level. Serbia is Serbia, so you can't use some existing code. And you also can't map stuff to a different country. But this can't be the first time somebody encountered this. Shouldn't there be someone at the tax office to ask?

But taking a step back: Serbia is not (yet) a member of the European Union, are you sure @thomaspaulb that you actually need to declare trade with Serbia? No intrastat document I can find says something about Serbia...

thomaspaulb:
I didnt give this much (enough) thought, i added Serbia because otherwise
the XAF export stops with an error on the records that refer to Serbia. So
either the XAF export should ignore all records from Serbia, or the spec
needs an update. I dont think it has anything to do with intrastat, since
also countries like Japan and Russia are in the list.

I looked at the specification document and the spec just says "country
code", iso 3166, 2 chars. So i would say that the tax office accepts any
such iso codes and the XSD they have supplied is just old/incomplete. But i
dont know for sure. I also looked quickly for the existence of "catch all"
codes, but couldnt find any thus far.

Op 2 mrt. 2017 16:10 schreef "Holger Brunn" <notifications@github.com>:

StefanRijnhart (approved):
Given that the XAF specs on this point are derived from ISO3166, and RS is included in ISO3166, I think in this case it is fine to fix the XAF specs.

astirpe (approved):
Currency CSD (Serbian dinar) is already contained in the XAF specs, so I'm definitely convinced that Serbia must be added to the list of countries :)

hbrunn:
I'll go with the majority here, even though I consider this dangerous because a file with this country simply is invalid according to the only official XSD I know

PR #63, opened 2017-03-07
[10.0][MIG] Port l10n_nl_tax_declaration_reporting

astirpe:
Porting module `l10n_nl_tax_declaration_reporting` to V10.
This is a work in progress...


astirpe:
Rebased PR, in order to apply the latest updates made on `.travis.yml`

l10n_nl_tax_declaration_reporting/report/report_vat.py
```diff
@@ -0,0 +1,161 @@
...
+            SELECT inv_tax.tax_id,
+                   sum(inv_tax.amount),
+                   sum(inv_line.price_subtotal_signed)
+            FROM account_invoice_tax AS inv_tax
+                JOIN account_invoice_line_tax inv_line_tax
+                    on inv_line_tax.tax_id = inv_tax.tax_id
+                JOIN account_invoice_line inv_line
+                    on inv_line.id = inv_line_tax.invoice_line_id
+                JOIN account_invoice inv
+                    on inv.id = inv_line.invoice_id
```
StefanRijnhart, line 151: AND inv.id = inv_tax.invoice_id

StefanRijnhart:
I see that the current version of the report does not include taxes posted directly on move lines. Is that the WIP-part?

astirpe:
@StefanRijnhart As far as I know, the VAT reporting should be only calculated based on invoices. But since I'm not accountant I'm going to ask to some financial consultant :)

**Thank you** for the tip on the query, I will update it ASAP!

And yes, I consider this PR still in an early WIP stage...

StefanRijnhart:
That assumption is just wrong.

But I think I would prefer a different approach altogether. Maybe this module could be refactored to reformat the results of https://github.com/OCA/account-financial-reporting/tree/10.0/account_tax_balance?

astirpe:
@StefanRijnhart Indeed it would be a better approach! Thank you for mentioning that module, I will work on this PR tomorrow...

l10n_nl_tax_declaration_reporting/report/report_vat.py
hbrunn: in my version for 9, I changed the query entirely, because the way the query stands here, it will return the amounts multiplied by the amount of lines which have the same tax (note that inv_tax.amount is the sum of *all* lines with the tax in the invoice, not just the amount of some specific line): https://github.com/OCA/l10n-netherlands/pull/68#issuecomment-285956281

StefanRijnhart:
@astirpe that would be great! You might want to check this PR of mine to fix the reporting of taxes on bank transactions in this module https://github.com/OCA/account-financial-reporting/pull/284

astirpe:
If the Community agrees, this PR is superseded by https://github.com/OCA/l10n-netherlands/pull/70

PR #66, opened 2017-03-10
[9.0][MIG] Port l10n_nl_xaf_auditfile_export

astirpe:
Based on the job made with #59


CasVissers-360ERP:
I tested this module on Odoo 9 Enterprise, it worked correctly. The only improvement I suggest is to add spaces between the data error warnings.

astirpe:
@CasVissers thank you! I agree. Could you check the spacing between errors with this latest commit?

CasVissers-360ERP:
@astirpe Thank you, this looks more clean and readable for the end user.

PR #70, opened 2017-03-18
[RFC][ADD] l10n nl tax statement for V10

astirpe:
# Introduction

Following the proposal of @StefanRijnhart discussed in PR https://github.com/OCA/l10n-netherlands/pull/63, I would have to rewrite from scratch the module `l10n_nl_tax_declaration_reporting` in order to port it from V8 to V10. The proposal was to use an existing OCA module `account_tax_balance` as a basis for retrieving the tax lines to be included in the Dutch tax statement report. Indeed the module  `account_tax_balance` does the job in an excellent way.

Here it is a proposal to implement a new module, that will replace `l10n_nl_tax_declaration_reporting`, introducing a different approach to generate and manage the BTW statements.

## Why change the existing V8 approach

- The Dutch BTW report on Odoo 8.0 uses the concept of Tax Code which is removed in Odoo V9 and later versions
- Starting from Odoo V9, we base the Tax reports on the Tags concept; this implies a complete redesign of the module `l10n_nl_tax_declaration_reporting` being ported to V9 and V10

## Status of the actual V8 module

- The module `l10n_nl_tax_declaration_reporting` allows the user to create PDF documents containing the Dutch tax statement report based on a date range; the module doesn't provide anything else
- A user who wants to just have a quick look to the report needs to generate and download a new PDF document; each time data in the system is added or changed, or just simply modifying the date range parameters, a new PDF document should be created and downloaded.

## Analysis of needed functionalities

- Would be nice a more handy and quick way to create and analyze the report, eg.: no need of creating a PDF document for just a quick analysis of the data
- If a tax statement is created and sent to the Dutch Taxation Authority, It would be nice to have that statement stored in the system for future availability
- If a tax statement is sent to the Dutch Taxation Authority and stored in the system, the document must be not changed nor updated; a concept of Status of the document (the 'statement' document) should be introduced
- Adding the Status to the statements, introducing the concept of Workflow is consequential: a document can be modified in Draft, cannot be modified is already Posted
- At the end the posted document should be transfered to the Dutch Taxation Authority that accepts the document in an electronic format (SBR/XBLR)
- Some other accounting systems other than Odoo allow to include in the report being created, not only the ordinary data filtered by the selected date range, but also all the old data that was not included in the previous tax declarations; this is an useful feature that could be included in the new module

## Features of the proposed new module

- The new proposed module allows an accountant to create a Tax Statement document
- The Tax Statement document accepts in input a date range and produces the report according with the Taxation Authority accepted format
- The module uses the existing OCA module `account_tax_balance` as a base reporting engine
- The calculation of the report is based on Tags set in the chart of tax accounts, like the default Dutch localization module containing the chart of accounts (module l10n_nl) which adds these tags automatically; the new module should be compliant with every chart of account having tags correctly configured
- The Tax Statement document can be set to two (or more) states: the document in Draft can be updated, the document in Posted status cannot
- The new module should be ready and open to every other improvements, so the idea is to keep it basic and simple for being extended easily

## Future developments

- The new module is predisposed for future implementation, eg.: exporting the tax statement in an  SBR/XBLR format (see proposal https://github.com/OCA/edi/issues/13, actually under development by Onestein)
- Include in the report being created, not only the ordinary data filtered by the selected date range, but also all the old data that was not included in the previous tax declarations (work in progress...)

## Request for Comments

Included with this RFC, I propose the module I'm actually developing.

Looking forward to read your comments!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 cache: pip
 
 addons:
+  postgresql: "9.2" # minimal postgresql version for the daterange method
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
--- a/l10n_nl_tax_statement/README.rst
+++ b/l10n_nl_tax_statement/README.rst
@@ -11,7 +11,8 @@ This module provides you with the Tax Statement in the Dutch format.
 Installation
 ============
 
-This module depends from module account_tax_balance available at https://github.com/OCA/account-financial-reporting.
+* This module depends from module account_tax_balance available at https://github.com/OCA/account-financial-reporting.
+* This module also depends from module date_range that is supported as of postgresql 9.2 or later versions.
 
 Configuration
 =============
--- a/l10n_nl_tax_statement/README.rst
+++ b/l10n_nl_tax_statement/README.rst
@@ -28,6 +28,7 @@ After that, go to go to menu: Invoicing -> Configuration -> Accounting -> NL BTW
 Usage
 =====
 
+#. Verify that you have enough permits. You need to belong at least to the Accountant group.
 #. Go to the menu: Invoicing -> Reports > Taxes Balance > NL BTW Statement
 #. Create a statement, providing a name and specifying start date and end date
 #. Press the Update button to calculate the report: the report lines will be displayed in the tab Statement
--- a/l10n_nl_tax_statement/README.rst
+++ b/l10n_nl_tax_statement/README.rst
@@ -25,7 +25,7 @@ Usage
 
 #. Go to the menu: Reports > Taxes Balance > NL BTW Statement
 #. Create a statement, providing a name and specifying start date and end date
-#. Press the Recompute button to calculate the report: the report lines will be displayed in the tab Statement
+#. Press the Update button to calculate the report: the report lines will be displayed in the tab Statement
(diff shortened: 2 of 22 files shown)


jvOrsouw:
@astirpe Nice job! the module looks really nice and I enjoy the 'post' feature!

- I noticed that tax 'l10n_nl.1_btw_ink_0_2' (BTW af te dragen verlegd (inkopen2)) with tag 'l10n_nl.tag_nl_34' (Voorbelasting (BTW) bis) was not shown in the report. The tax should be shown in categorie 5b.

astirpe:
@jvOrsouw thank you for the feedback! Should be fixed with the latest commit.

nielsmiddeldorp:
👍

jvOrsouw:
please correct the calculation of 5a: this should be a the sum of 1a through 4b.

jvOrsouw (approved):
I've tested the taxes with debit and credit values for 5a and 5b (and the categories they consist of), and found that the taxes ends up in the right category, with the right amounts and that the sum of the taxes is calculated correctly in all cases. 

👍

hbrunn:
@nielsmiddeldorp @jvOrsouw please use the github review workflow next time, but I count your two thumbs up as approval and merge without reviewing myself

PR #75, opened 2017-04-25
[FIX] Limit userID to maximum 35 characters

### LINKED ISSUE #71: [9.0] XAF audit file export fails with long login (E-Mail)
With the new 9.0 module, the export fails when the user doing the entries logs in with an emailadres.
The error message:

<pre>
:5228:0:ERROR:SCHEMASV:SCHEMAV_CVC_MAXLENGTH_VALID: Element '{http://www.auditfiles.nl/XAF/3.2}userID': [facet 'maxLength'] The value has a length of '37'; this exceeds the allowed maximum length of '35'.
:5228:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://www.auditfiles.nl/XAF/3.2}userID': 'p.pietjepuk@pietjepukenterprisesbv.nl' is not a valid value of the atomic type '{http://www.auditfiles.nl/XAF/3.2}IdentificationString35'.
</pre>

We are using google authentication, therefore we can't change the login.
Proposed solutions:
- Replace the login with the part before the @, if the login contains an @.
- Truncate the login to 35 characters.. This usually truncates the last part of the domain,
- Use the name, of another field?

If someone can give feedback, I can create a patch for this.

RoelAdriaans:
Fixes #71

### DIFF
--- a/l10n_nl_xaf_auditfile_export/__openerp__.py
+++ b/l10n_nl_xaf_auditfile_export/__openerp__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 {
     "name": "XAF auditfile export",
-    "version": "9.0.1.0.0",
+    "version": "9.0.1.0.1",
     "author": "Therp BV, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Accounting & Finance",
--- a/l10n_nl_xaf_auditfile_export/views/templates.xml
+++ b/l10n_nl_xaf_auditfile_export/views/templates.xml
@@ -65,7 +65,7 @@
                     <bankIdCd><t t-esc="bank.bank_bic or bank.bank_id.bic" /></bankIdCd>
                 </bankAccount>
                 <changeInfo t-if="p.write_uid">
-                    <userID><t t-esc="p.write_uid.login" /></userID>
+                    <userID><t t-esc="p.write_uid.login[:35]" /></userID>
                     <changeDateTime><t t-esc="p.write_date[:10]" />T<t t-esc="p.write_date[-8:]" /></changeDateTime>
                     <changeDescription>Last write</changeDescription>
                 </changeInfo>
@@ -94,7 +94,7 @@
                 </taxonomy>
                 /-->
                 <changeInfo t-if="a.write_uid">
-                    <userID><t t-esc="a.write_uid.login" /></userID>
+                    <userID><t t-esc="a.write_uid.login[:35]" /></userID>
                     <changeDateTime><t t-esc="a.write_date[:10]" />T<t t-esc="a.write_date[-8:]" /></changeDateTime>
                     <changeDescription>Last write</changeDescription>
                 </changeInfo>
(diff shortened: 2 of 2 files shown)


pankk:
Looks like this would work in practice, but in theory, is a unique userID required?
For example, if we have the following logins:
'pietjepukenterprisesholdingcompany01@something.com'
'pietjepukenterprisesholdingcompany02@something.com'
'pietjepukenterprisesholdingcompany03@something.com'
We'd truncate them all to just 'pietjepukenterprisesholdingcompany0'. Would this be a problem?

hbrunn (changes requested):
If you want to shorten fields, use a widget as the other fields do, you'll have to declare a new one in https://github.com/OCA/l10n-netherlands/blob/9.0/l10n_nl_xaf_auditfile_export/models/ir_qweb.py

But given we only need an ID here, why not use the user's ID here anyways? This would also address @pankk's concern

RoelAdriaans:
@hbrunn I tried making a widget, but it didn't work. See my commit: https://github.com/RoelAdriaans-B-informed/l10n-netherlands/commit/c5160e8dc1fa97170aa10e259a1c17804ea92ea8 , still fails.
But, I think that's an issue with the widget, debugging shows that _format() is never called... 

Indeed, @pankk has a valid point with the usernames, that's why I asked for feedback in #71 . No problems with using an ID here, standard makes this possible ( http://www.ictplaza.nl/uploads/xml_auditfiles/xmlfinancieel/20140402_AuditfileFinancieelVersie_3_2.zip , file XmlAuditfileFinancieel3.2_FunHie.rtf ). It also defines a basics table linking the userid to the real username..

hbrunn:
the widgets not being used will be a problem anyways, this worked in 8.0 - that seems an oversight during the migration to 9.0. This will cause a lot of other issues too for too long character fields, so it's certainly something to look into

astirpe:
@RoelAdriaans-B-informed with commit https://github.com/astirpe/l10n-netherlands/commit/5f7d133ff739e3c4387834c546f8fca8550e5cde the widgets should work. Could you check?
I opened the pull request https://github.com/RoelAdriaans-B-informed/l10n-netherlands/pull/1 to your branch.

hbrunn:
fixed in d9931e2, closing this

astirpe:
@hbrunn I think this one is not fixed yet. Am I missing something?

hbrunn:
@astirpe yes, your own backport of the proper fix: https://github.com/OCA/l10n-netherlands/pull/94 ;-)

astirpe:
@hbrunn I see, but this PR says: "[FIX] Limit userID to maximum 35 characters".
And the userID length was not touched in the main repository.

Moreover, in this PR there is a discussion still pending, regarding the adoption of the ID of the user instead of the truncation of the user.login, as it was initially proposed.

hbrunn:
right, a bit premature, thanks. Then it just needs fixing

astirpe:
Thanks!

... and by the way I vote for the adoption of the user ID, instead of the actual user login.

astirpe (changes requested):
@RoelAdriaans-B-informed if you agree, you could replace `write_uid.login` with `write_uid.id`. If you also bump the version number, this PR is ready to be merged IMO.

RoelAdriaans:
I agree with the user id. Committed and bumped version

astirpe (approved):

PR #78, opened 2017-05-15
[IMP][8.0] Extract method in l10n_nl_xaf_auditfile_export

astirpe:
This PR is meant to simplify the way we can extend the auditfile template of module `l10n_nl_xaf_auditfile_export`.

Let's say I want to create a separate module to extend the auditfile template and, based on a certain condition, apply that extended template. For example, think about fixing issue #71 by creating a customized module.

In such cases, it would be very useful to have a simple method that can be easily extended, in order to retrieve the custom template.
This PR is a proposal to extract method `def _get_auditfile_template()` from the existing `def button_generate()`.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -92,10 +92,10 @@ class XafAuditfileExport(models.Model):
     @api.multi
     def button_generate(self):
         self.date_generated = fields.Datetime.now(self)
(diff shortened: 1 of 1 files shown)


hbrunn (approved):
from your code I understand that you need to pick a different template depending on values? Otherwise, I'd have suggested to make this a model method.

astirpe:
@hbrunn Yes, in my case the template is based on a field of the record. I need to read some values from the singleton `self`. That's why you notice the `@api.multi` decorator.

StefanRijnhart (approved):

PR #81, opened 2017-05-31
 [FIX][8.0] Auditfile: add RUB currency.

astirpe:
This is the V8 porting of PR #79.


thomaspaulb (approved):
Code looks good, and the rationale of this has been already discussed in PR #79, so I approve this fully.

Issue #23, opened 2015-09-09
CBS / Intrastat module for The Netherlands


andreschenkels:
Based on the OCA intrastat base we should provide a correct Intrastat report

hbrunn:
https://github.com/OCA/l10n-netherlands/pull/57

astirpe:
CBS and ICP: are we talking about different things?

- Centraal Bureau voor de Statistiek (CBS)
- Opgaaf intracommunautaire prestaties (ICP)

ploegvde:
Yes those are different things. CBS is for the Dutch Statistics Bureau and ICP is for the Dutch Tax authorities. Reports are completely different.

astirpe:
Are we considering to reopen this issue? Would CBS reports be of interest for the Dutch community?

ploegvde:
I think the would. WE have made something recently, The CBS output is a CSV file (yes they are that old fashioned) That you have to upload to the CBS website.

The only thing is that the content contains some fixed values, that are fixed for my customer, but can we different for other customer. But i can share if you want.

astirpe:
@ploegvde if you want, you can open a PR to OCA proposing that module, we will continue the discussion there. It would be great!
For making it generic and OCA guidelines compliant we can join the efforts.
What do you think?

ploegvde:
Ok, i will do that today.

PR #84, opened 2017-06-06
Added new module l10n_nl_cbs_export

ploegvde:
@astirpe  As discussed in issue #23. This is they CBS export module we use at one of our customers. The module creates a cron job and creates the export file automatically every month. We have added some users as followers of the CBS export so that they get notified.

The output file has to be manually uploaded on the CBS site. There is no other way to my knowledge.

In the module some values of fields are hard codes and need configuration, so that the module is suitable for other companies. I have no problem in doing that. 

So the question is, if the module is something we can use as a start?


astirpe:
@ploegvde thank you for this, indeed is a good start!
Before starting the review, could you work on it, in order to make the module more generic and configurable?
Let me know in case you need help.

ploegvde:
yes, i will do that. No problem.

l10n_nl_cbs_export/LICENSE
```diff
@@ -0,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
hbrunn: the licence key is enough imho

l10n_nl_cbs_export/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+        "view/cbs_export_file_sequence.xml",
+        "data/cron_process_cbs_export.xml",
+        "view/cbs_export_file_view.xml",
+    ],
+    'demo_xml': [],
+    'test': [],
+    'installable': True,
+    'auto_install': False,
+    'application': False,
+    'license': 'LGPL-3',
```
hbrunn: this is inconsistent with the license file, do you mean agpl-3 here?

ploegvde:
@astirpe Hi, i checked the module and the output. In my branch

https://github.com/Odoo-experts/l10n-netherlands/tree/l10n_nl_cbs_export

I have added documentation of CBS. It is only in Dutch.

What i see is that the fixed values we use are for most common situations correct. There is only 1 field that needs to be added to product template and that is "Bijzondere maatstaf"

But what i'm doubting about is the use of the "intrastat-code". In Odoo V10 we have now also the HS-Code and to me it is not clear what the difference is. So maybe others can help on this.

So think the module can get a fitst review based on what it does now?

l10n_nl_cbs_export/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+        See also: https://www.cbs.nl/nl-nl/deelnemers-enquetes/deelnemers-enquetes/bedrijven/onderzoek/lopend/internationale-handel-in-goederen/idep-codelijsten
+        
+    """,
+    "data": [
+        "security/ir.model.access.csv",
+        "view/cbs_export_file_sequence.xml",
+        "data/cron_process_cbs_export.xml",
+        "view/cbs_export_file_view.xml",
+    ],
+    'demo_xml': [],
```
hbrunn: remove empty keys and ones with the defaul value

l10n_nl_cbs_export/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "website": "https://www.odooexperts.nl",
+    "category": "Accounting",
+    "depends": ["account", "report_intrastat"],
+    "summary": "CBS Export File for Dutch Intrahandel Sale",
+    "description": """
+        CBS Export File for Dutch Intrahandel Sale
+        
+        See also: https://www.cbs.nl/nl-nl/deelnemers-enquetes/deelnemers-enquetes/bedrijven/onderzoek/lopend/internationale-handel-in-goederen/idep-codelijsten
+        
+    """,
```
hbrunn: remove this and use a [README.rst](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)

l10n_nl_cbs_export/data/cron_process_cbs_export.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" ?>
+<openerp>
+    <data noupdate="1">
+        <record id="cbs_export_file" model="ir.cron">
+            <field eval="&quot;&quot;&quot;cron_get_cbs_export_file&quot;&quot;&quot;" name="function"/>
```
hbrunn: you don't need any of those evals, replace them by the plain value

l10n_nl_cbs_export/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# See LICENSE file for full copyright and licensing details.
+
+import account_invoice
+import cbs_export_file
```
hbrunn: use relative imports

l10n_nl_cbs_export/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# See LICENSE file for full copyright and licensing details.
+import models
```
hbrunn: use relative imports

l10n_nl_cbs_export/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# See LICENSE file for full copyright and licensing details.
+
+{
+    "name": "CBS Export Intrahandel Sale",
+    "version": "1.0",
+    "author": "Odoo Experts",
```
hbrunn: add OCA as first author

l10n_nl_cbs_export/models/cbs_export_file.py
```diff
@@ -0,0 +1,146 @@
...
+        else:
+            account_ids.search([('cbs_export_id', '=', self.id)]).write(
+                {'cbs_export_id': False}
+            )
+            account_ids.write({'cbs_export_id': self.id})
+
+    @api.model
+    def cron_get_cbs_export_file(self):
+        last_month = date.today().replace(day=1) - timedelta(days=1)
+        sr_ids = self.search([('month', '=', last_month.strftime("%m")), ('year', '=', last_month.strftime("%Y"))])
```
hbrunn: this is not a list of ids, but a recordset

l10n_nl_cbs_export/models/cbs_export_file.py
```diff
@@ -0,0 +1,146 @@
...
+                is_valid_year = '%d' % (int(self.year))
+                try:
+                    time.strptime(is_valid_year, '%Y')
+                except ValueError:
+                    raise ValidationError(_("Please insert a valid Year"))
+
+    @api.one
+    def set_invoice(self):
+        days = calendar.monthrange(int(self.year), int(self.month))
+        account_ids = self.env['account.invoice'].search([
```
hbrunn: `account_ids` is a very misleading name for a variable holding a recordset of invoices

l10n_nl_cbs_export/models/cbs_export_file.py
```diff
@@ -0,0 +1,146 @@
...
+                subtype='mt_comment'
+            )
+        sr_ids.set_invoice()
+        sr_ids.export_file()
+
+    @api.one
+    def export_file(self):
+        line_number = 1
+        company_id = self.env.user.company_id
+        cbs_export_data = str('9801')+str(
```
hbrunn: replace all of this by a qweb template that does all the formatting etc (compare with the xaf export)

hbrunn (changes requested):
There are also a lot of complaints from runbot: https://travis-ci.org/OCA/l10n-netherlands/jobs/240298118#L556-L580

I think you can save yourself some inconvenience by using a date field or (if the reporting period can be something else than months) date ranges. Then you save those files and jump through some hoops to make this unique (shouldn't there be a uniqueness constraint for good measure?), is there a reason to do this or can't you just generate those on the fly as requested?

astirpe:
@ploegvde great, thanks! I will start a review as soon as I have some time.

There is also an OCA module [product_harmonized_system](https://github.com/OCA/intrastat/tree/10.0/product_harmonized_system) that handles the HS codes.

l10n_nl_cbs_export/view/cbs_export_file_view.xml
```diff
@@ -0,0 +1,87 @@
...
+                    <group>
+                        <field name='name' invisible='True'/>
+                        <field name="cbs_export_invoice" filename='filename' attrs="{'invisible': [('cbs_export_invoice', '=', False)]}"/>
+                    </group>
+                    <group>
+                        <field name='month' attrs="{'readonly': [('cbs_export_invoice', '!=', False)]}"/>
+                        <field name='year' required='True' attrs="{'readonly': [('cbs_export_invoice', '!=', False)]}"/>
+                    </group>
+                  <notebook>
+                        <page string="account Invoice">
```
astirpe: Instead of "account Invoice", this tab could be named in a more user friendly way, eg.: "Invoices".

l10n_nl_cbs_export/view/cbs_export_file_view.xml
```diff
@@ -0,0 +1,87 @@
...
+        </field>
+    </record>
+
+    <record id="view_cbs_export_file_form" model="ir.ui.view">
+        <field name="name">cbs.export.file.form</field>
+        <field name="model">cbs.export.file</field>
+        <field name="arch" type="xml">
+            <form string="CBS Export" >
+                <header>
+                    <button name="get_data" string="CBS Export Again" type="object" class="oe_highlight" attrs="{'invisible': [('cbs_export_invoice', '!=', False)]}"/>
```
astirpe: I manually created a record of model `cbs.export.file` without any export file. The button "CBS Export Again" remains visible, I was expecting it to be hidden because of this:
`
attrs="{'invisible': [('cbs_export_invoice', '!=', False)]}"
`
so it seems to me that the invisible attribute is not working.

However I think you can remove the invisible attribute at all, so that the button will be always displayed. In this case just rename the button "CBS Export Again" to something like "Manual CBS Export".

You can also display the button only for a certain group (eg.: Account Advisor) by adding a `groups=" ... "` definition.

l10n_nl_cbs_export/models/cbs_export_file.py
ploegvde: Why would this be needed?

hbrunn: it gives other developers a modular way to hook into the generation process and modify it. So it's needed to make your code better.

ploegvde: I would like the module first to be ready and functional. If it needs improvement, then we improve later. I like users to benefit of what now is available. Is that possible?

astirpe: Considering that this is the first @ploegvde 's contribution to the OCA repository (if I'm not wrong) and in the spirit of the OCA principles, I think that the proposal of Erwin can be accepted. I will convert this formatting part to a qweb template later on, after this PR is merged, in case nobody wants to do it.

l10n_nl_cbs_export/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
```
astirpe, line 1: I don't see any group with creation rights for model cbs.export.file.
At least group "Account Advisor" must have creation/write rights.

astirpe (changes requested):

hbrunn (approved):

astirpe (approved):

ploegvde:
Can this be merged?

PR #88, opened 2017-06-21
[FIX][10.0] NL BTW statement, onchange date range

### LINKED ISSUE #87: NL BTW statement, date range not working
When creating a new NL BTW Statement I select a date range. Unfortunately the date range is not used in de two date fields in the NL BTW Statement form.

astirpe:
Fixes #87

### DIFF
--- a/l10n_nl_tax_statement/models/l10n_nl_vat_statement.py
+++ b/l10n_nl_tax_statement/models/l10n_nl_vat_statement.py
@@ -72,8 +72,10 @@ class VatStatement(models.Model):
     @api.onchange('date_range_id')
     def onchange_date_range_id(self):
         if self.date_range_id and self.state == 'draft':
-            self.from_date = self.date_range_id.date_start
-            self.to_date = self.date_range_id.date_end
+            self.update({
+                'from_date': self.date_range_id.date_start,
+                'to_date': self.date_range_id.date_end,
+            })
 
     @api.onchange('from_date', 'to_date')
     def onchange_date(self):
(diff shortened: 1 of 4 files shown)


l10n_nl_tax_statement/models/l10n_nl_vat_statement.py
```diff
@@ -72,10 +72,8 @@ def default_get(self, fields_list):
     @api.onchange('date_range_id')
     def onchange_date_range_id(self):
         if self.date_range_id and self.state == 'draft':
-            self.write({
-                'from_date': self.date_range_id.date_start,
-                'to_date': self.date_range_id.date_end,
-            })
+            self.from_date = self.date_range_id.date_start
+            self.to_date = self.date_range_id.date_end
```
hbrunn: this is just fine too, but are you aware you can use the [`.update`](https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L5015) function for exactly those cases? A lot more convenient for bigger amounts of fields

astirpe: I didn't now, thank you!

You just found another issue: on this module I created a method called `update()` that does something else, but actually it overrides the standard `update()`. I'm fixing it.

hbrunn: `found` might be an exaggeration, and shame on me for not noticing during review...

CasVissers-360ERP (approved):
Tested

hbrunn (approved):

PR #89, opened 2017-06-21
[FIX][9.0] NL BTW statement, onchange date range

### LINKED ISSUE #87: NL BTW statement, date range not working

astirpe:
Fixes #87

### DIFF
--- a/l10n_nl_tax_statement/views/l10n_nl_vat_statement.xml
+++ b/l10n_nl_tax_statement/views/l10n_nl_vat_statement.xml
@@ -11,7 +11,7 @@
                 </header>
                 <sheet>
                     <div class="oe_button_box" name="button_box">
-                        <button name="update" string="Update" states="draft" type="object" class="oe_stat_button" icon="fa-repeat"/>
+                        <button name="statement_update" string="Update" states="draft" type="object" class="oe_stat_button" icon="fa-repeat"/>
                         <button name="post" string="Post" states="draft" type="object" class="oe_stat_button" icon="fa-arrow-right text-success"/>
                     </div>
                     <label for="name"/>
--- a/l10n_nl_tax_statement/views/l10n_nl_vat_statement.xml
+++ b/l10n_nl_tax_statement/views/l10n_nl_vat_statement.xml
@@ -21,7 +21,7 @@
                     <group>
                         <group name="tax_report">
                             <field name="company_id" options="{'no_create': True}" groups="base.group_multi_company"/>
-                            <field name="date_range_id" domain="[('company_id','=',company_id)]" attrs="{'readonly': [('state','=','posted')]}"/>
+                            <field name="date_range_id" attrs="{'readonly': [('state','=','posted')]}"/>
                             <field name="from_date" attrs="{'readonly': [('state','=','posted')]}"/>
                             <field name="to_date" attrs="{'readonly': [('state','=','posted')]}"/>
                         </group>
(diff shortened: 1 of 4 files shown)


astirpe:
Runbot failure not related with this PR:

```
2017-06-21 07:40:07,624 174 ERROR openerp_test openerp.modules.graph: module l10n_nl_postcodeapi: Unmet dependencies: partner_street_number

```

It is because the porting of `partner_street_number` module is still pending. See PR: OCA/partner-contact#207

hbrunn (approved):

astirpe:
I'm also including the backport of the fix https://github.com/OCA/l10n-netherlands/pull/97

CasVissers-360ERP (approved):

astirpe:
This PR got 2 approvals long time ago. The reason of the runbot failure is described here https://github.com/OCA/l10n-netherlands/pull/89#issuecomment-309994329. Every new PR on branch 9.0 will fail for that reason.

So in my opinion this PR can be merged.

hbrunn:
we'll get an inconsistent repo this way. If you want to speed this up, I'd suggest taking over the partner_street_number branch, fix the last comments and increase test coverage. Then we can merge there and afterwards here

astirpe:
The repo (branch 9.0) is already inconsistent. The cause is not related with this PR. The cause is related with the module l10n_nl_postcodeapi, that was merged with broken dependencies long time ago. I don't want to speed up this PR, but I don't see a valid reason to block an obvious back-porting of the bug fixing that was already accepted and merged in branch 10.

l10n_nl_tax_statement/views/l10n_nl_vat_statement.xml
```diff
@@ -21,7 +21,7 @@
                     <group>
                         <group name="tax_report">
                             <field name="company_id" options="{'no_create': True}" groups="base.group_multi_company"/>
-                            <field name="date_range_id" domain="[('company_id','=',company_id)]" attrs="{'readonly': [('state','=','posted')]}"/>
```
StefanRijnhart, line 24: I believe company_id is not required on date ranges, so you might want to allow for ranges without a company here.

l10n_nl_tax_statement/views/l10n_nl_vat_statement.xml
```diff
@@ -21,7 +21,7 @@
...
+                            <field name="date_range_id" attrs="{'readonly': [('state','=','posted')]}"/>
```
astirpe, line 24: This line fixes https://github.com/OCA/l10n-netherlands/issues/96 for V9

l10n_nl_tax_statement/views/l10n_nl_vat_statement.xml
astirpe: Yes, true. That's why I'm removing the domain in this commit 576473fdffae884d6902ffafc4cf788975a96f5c

StefanRijnhart: LOL, sorry!

StefanRijnhart (approved):

StefanRijnhart:
Agree with @astirpe that this does not break anything that is not already broken, so could be merged IMHO.

hbrunn:
agreed, here i overdid it with my rule fetishism (which I consider a good thing in general). Sorry @astirpe and @StefanRijnhart, thanks for making this clear

PR #90, opened 2017-06-21
[FIX][10.0] l10n_nl_intrastat: onchange date range

astirpe:
Fixes method `onchange_date_range_id()` on module `l10n_nl_intrastat`, the same way as it was fixed in #88


hbrunn (approved):
we'll squash during merge

CasVissers-360ERP (approved):
Tested

PR #93, opened 2017-07-03
[FIX][10.0] Missing multicompany security rule

astirpe:
In module `l10n_nl_tax_statement` the multicompany security rule was missing.


CasVissers-360ERP (approved):
Tested & approved

PR #101, opened 2017-07-25
[FIX] don't cap utf8 strings, add test data for that

hbrunn:
for shortening strings exceeding the limits given in the schema, we can't slice the utf8 version, as this then only looks at bytes. If a multi byte unicode character happens to be the last one, it will be cut in the middle, yielding invalid utf8 and thus a broken audit file. Try the test data without the patch to see the effect.

@astirpe I bet you'll want to cherry pick this to 9 and 10

### DIFF
--- a/l10n_nl_xaf_auditfile_export/__openerp__.py
+++ b/l10n_nl_xaf_auditfile_export/__openerp__.py
@@ -30,6 +30,7 @@
         'account',
     ],
     "data": [
+        "demo/res_partner.xml",
         "views/xaf_auditfile_export.xml",
         "views/menu.xml",
         'views/templates.xml',
--- a/l10n_nl_xaf_auditfile_export/demo/res_partner.xml
+++ b/l10n_nl_xaf_auditfile_export/demo/res_partner.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <record id="demo_partner" model="res.partner">
+            <field name="name">Demo partner with special character in zip at cutoff position</field>
+            <field name="zip">123456789Ω</field>
+        </record>
+        <record id="base.main_partner" model="res.partner">
+            <!-- it's not important this this is NL, but it must be something /-->
+            <field name="country_id" ref="base.nl" />
+        </record>
+    </data>
+</openerp>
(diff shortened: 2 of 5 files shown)


astirpe (approved):
@hbrunn  Thanks! Did you have problems with any customer regarding this particular case? How did you discover this issue?

hbrunn:
this was a Swedish customer who had an umlaut at this position (and misspelled the zip code anyways ;-) )

luc-demeyer:
I missed this PR when creating my PR #115 (I didn't find a reference to XAF in the PR description and I am not following the l10n-nl branch actively).

I have now reviewed the code and tested this PR.
It fixes the issue and generates a XAF file which is identical to the one generated with my PR.
The 'needs review' label can as a consequence be removed and 👍  for merge.

Technically I prefer a little bit the mix of both PR's:
In my approach I avoid unnecessary unicode to UTF-8 conversions.
I perform the truncate of the unicode string before the qweb engine converts it to UTF-8, hence no need to convert the qweb rendered UTF-8 string back to unicode for the truncate and than again to UTF-8. 

If you agree with this alternative truncate logic, than I'll make a PR on top of this PR.

hbrunn:
good point about the PR description, will try to remember to add more context in the future.

I'm all for avoiding unnecessary conversions, so please go ahead. But I'd like to keep calling `eval_str` instead of cooking up our own version. How does your version cap in the unicode case btw?

How about: Call `eval_str`, check if any of the bytes there has the first bit set, in that case, do the unicode dance, otherwise, cap the utf8 version?

luc-demeyer:
I don't think it's a big deal to have the '_trunk_eval' as a replacement for the 'eval_str' method since it's only a few lines of code and the sensitive work (the save_eval via the ir.qweb eval method) is kept.
Checking the first bit set means again a CPU intensive string parsing operation on potentially millions of transaction lines (I made the fix for a XAF auditfile on a spanish database, nearly every word in Spain has special characters).

hbrunn:
please add a second commit to fix https://travis-ci.org/OCA/l10n-netherlands/jobs/275206966#L739, then we can fasttrack this

hbrunn:
just noticed this is my own branch. Fixed, and counting @luc-demeyer's thumbs up as an approval, we can merge

PR #108, opened 2017-08-17
[ADD][10.0] Module l10n_nl_tax_invoice_basis (Factuurstelsel)

astirpe:
Following the proposal (see [this comment](https://github.com/odoo/odoo/issues/18572#issuecomment-322170029)) to add the Factuurstelsel in the Dutch localization for Odoo 11, I'm proposing here a module for making the Odoo 10.0 accounting system compliant with the Factuurstelsel as well.

@yung-wang I look forward to read your feedbacks, since this topic was your original proposal!


yung-wang:
@astirpe I made 2 invoice both with invoice date 18-aug. 1 invoice has accounting date 01-jul. When I create the VAT report "Aangifte omzetbelasting" it shows the correct VAT amount in the right daterange.
But in the report Taxes Balance, the Base Balance is stil taking the accounting date 01-jul. Can you have a check if that can be fixed too?

astirpe:
@yung-wang yes, you are right! The Base Balance is actually misaligned in the report, I'm looking at it.

Thank you!

astirpe:
I've added a dependency to `account_tax_balance`, in order to fix the report problem
@yung-wang May I ask you to take a look at it?

yung-wang (approved):
Hi Andrea, I checked the new commits and they are good.

I create invoices with August as invoice date and same accounting date. Also some invoice with a different accounting date.
Further validate some other invoices and did some credit invoices too.

The standard reports like Balance, P&L are looking fine.
Both Taxes Balance and Dutch TAX report shows the correct VAT + base correctly according the Factuurstelsel (the invoice date is leading).

yung-wang:
To make the Factuurstelsel complete, I think it is a good idea to add Vat Report posted information in the move lines. To do a VAT report the tax authority looks at the invoice date. Let's say you already did and sent the VAT report of July. Suddenly you find in the month August another invoice. According the law you're allowed to report the VAT of forgotten invoice in month August.

So in Report Jul, you have only invoices of Jul. In August you have the forgotten invoices of Jul + Aug. It would also be a nice feature, that the system generates a report that shows these 'forgotten' invoices.
Also see if it is possible to mark a VAT report as final. Once it is final, you cannot change it. If you  unpost a posted VAT report, all the moves lines that have been posted in that month should be set open for VAT report posting.

Hope you can improve this function.

l10n_nl_tax_invoice_basis/models/account_tax.py
```diff
@@ -0,0 +1,33 @@
...
+                ('tax_date', '!=', False),
+                '&',
+                ('tax_date', '<=', to_date),
+                ('tax_date', '>=', from_date),
+                '&',
+                ('tax_date', '=', False),
+                '&',
+                ('date', '<=', to_date),
+                ('date', '>=', from_date),
+            ]
```
espo-tony: Though this domain list works properly, it is rather hard to read and surely not understandable at a first glance.
Can I suggest you to split it in order to be more readable? Something like this could already be more readable:

```
tax_date_condition = ['&', '&', ('tax_date', '!=', False), ('tax_date', '<=', to_date), ('tax_date', '>=', from_date)]
date_condition = ['&', '&', ('tax_date', '=', False), ('date', '<=', to_date), ('date', '>=', from_date)]
return [('company_id', '=', company_id), '|',] + tax_date_condition + date_condition
```

In addition, it could also be useful to add a comment to explain what the condition means, either in natural language or at least using a much more readable infix notation.

espo-tony (changes requested):

astirpe:
@yung-wang I like the idea to make the Factuurstelsel complete, thank you! I'm going to work on it.

Even if your proposal is part of the Factuurstelsel functionality, the solution will be implemented in the other module `l10n_nl_tax_statement` instead. I don't think there is much to add in this pull request in order to improve the workflow of the VAT report. What do you think?

l10n_nl_tax_invoice_basis/models/account_tax.py
astirpe: Even better, we could split the definition of `tax_date_condition` and `date_condition` in separate methods.

yung-wang:
@astirpe That is a good proposal. Let's continue and make this complete in the other modules. Thanks, this pull request is completed.

espo-tony (approved):
Thanks. LGTM

PR #111, opened 2017-08-21
[FIX][8.0] l10n_nl_xaf_auditfile_export: export periodNumber

astirpe:
On an old multi-company database with a lot of companies we ran into this:
```
:21714:0:ERROR:SCHEMASV:SCHEMAV_CVC_TOTALDIGITS_VALID: Element '{http://www.auditfiles.nl/XAF/3.2}periodNumber': [facet 'totalDigits'] The value '3167' has more digits than are allowed ('3'). :21714:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://www.auditfiles.nl/XAF/3.2}periodNumber': '3167' is not a valid value of the atomic type '{http://www.auditfiles.nl/XAF/3.2}Nonnegativeinteger3'. :21719:0:ERROR:SCHEMASV:SCHEMAV_CVC_TOTALDIGITS_VALID: Element 

```
The issue is that the ID of `account_period` (for periodNumber) has more that three digits. This is not allowed as per the XAF Auditfile schema.

As I already proposed for version 10.0 (see https://github.com/OCA/l10n-netherlands/pull/59#issuecomment-282017404), I'm proposing to export the period numbers (`periodNumber`) using 3 digits: 1st digit is the last digit of the fiscal year, the other 2 digits are the month (Eg.: period "August 2017" is represented by number 708).

So this is the backport to 8.0 of the periodNumber representation already made for 10.0.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/views/templates.xml
+++ b/l10n_nl_xaf_auditfile_export/views/templates.xml
@@ -114,7 +114,7 @@
         </vatCodes>
         <periods>
             <period t-foreach="self.get_periods()" t-as="p">
-                <periodNumber><t t-esc="p.id" /></periodNumber>
+                <periodNumber><t t-esc="p.code[-1:] + p.code[0:2]" /></periodNumber>
                 <periodDesc><t t-esc="p.name" t-esc-options='{"widget": "auditfile.string50"}' /></periodDesc>
                 <startDatePeriod><t t-esc="p.date_start" /></startDatePeriod>
                 <endDatePeriod><t t-esc="p.date_stop" /></endDatePeriod>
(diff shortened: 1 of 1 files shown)


hbrunn (approved):

astirpe:
@hbrunn for this PR, the milestone should be set to 8.0 (and not to 10.0)

ThijsvOers (approved):
I have tested the fix Andrea made in a 8.0 environment and it works fine.

luc-demeyer:
I intended to rebase my PR https://github.com/OCA/l10n-netherlands/pull/119 because it conflicts with this change that got merged.
Before I can do this rebase, this change in this merge must be reverted and fixed because this solution is wrong.
Let me explain:

The period code field is a field that can be entered/modfied by the end user.
We do know that the standard accounting wizard generates period codes via the logic 'mm/yyyy'.
But users can change this and I have seen many installs where the periods are called 'yyyy-mm' (makes actually more sense than the standard way for period based sorting purposes).

I think we should keep the logic proposed by andrea but implemented properly (take the month and year of the period date_stop field in order to compose periodNumber in the XAF file.

astirpe:
@luc-demeyer is ok for me. Will you fix it in your PR?

luc-demeyer:
cf. https://github.com/OCA/l10n-netherlands/pull/149
As soon as this one is merged I'll rebase my PR 119.

PR #115, opened 2017-09-13
fix xaf auditfile

luc-demeyer:
The xsd trunc should take place before the utf-8 conversion..
If not, the string truncate may end up with a partial special character resulting in a dump in the lxml iterparse function (e.g. char \xc3\x81 truncated into \xc3 which is an incorrect utf-8 char).


hbrunn:
that's a quite complicated version of https://github.com/OCA/l10n-netherlands/pull/101 which also adds tests, so please review & merge that

PR #119, opened 2017-10-03
[8.0] fix xaf auditfile contents

luc-demeyer:
This PR fixes the XAF query to ensure that the partners, accounts and journals used in the selected periods are also included in the XAF file.
On the other hand partners, accounts and journals NOT involved in transactions in the selected periods are not relevant for the auditors and should not be included.
The level of detail of the export can also be selected. By default this level is limited to what required by external auditors since it's not a normal practice to e.g. give your customer database details to an external party.

This PR also includes PR https://github.com/OCA/l10n-netherlands/pull/101 (this fix really should be merged asap so that a XAF export no longer ends up in a python dump).

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -150,7 +150,7 @@ class XafAuditfileExport(models.Model):
             "WHERE period_id IN %s AND account_id NOT IN %s "
             "AND company_id=%s ",
             (tuple(self._periods._ids),
-             tuple(self.exclude_account_ids._ids),
+             tuple(self.exclude_account_ids._ids or [0]),
              self.company_id.id))
         res = self._cr.fetchall()
 
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -90,9 +90,13 @@ class XafAuditfileExport(models.Model):
     @api.multi
     def button_generate(self):
         self.date_generated = fields.Datetime.now(self)
-        self._get_data()
+        accounts, journals, partner_ids, periods = self._get_data()
         auditfile_template = self._get_auditfile_template()
         xml = auditfile_template.render(values={
+            'accounts': accounts,
+            'journals': journals,
+            'partner_ids': partner_ids,
+            'periods': periods,
             'self': self,
         })
         # the following is dealing with the fact that qweb templates don't like
@@ -138,7 +142,7 @@ class XafAuditfileExport(models.Model):
         )
 
     def _get_data(self):
-        self._periods = self.env['account.period'].search([
+        periods = self.env['account.period'].search([
             ('date_start', '<=', self.period_end.date_stop),
             ('date_stop', '>=', self.period_start.date_start),
             ('company_id', '=', self.company_id.id),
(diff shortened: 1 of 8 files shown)


hbrunn:
is there a chance to have the all template inherit from the default template to avoid code duplication?

and next time you have to include another branch's code in your branch, better cherry pick those commits. When we merge #101, this will conflict and then you have to revisit this PR again.

hbrunn:
now please rebase

luc-demeyer:
rebased.

hbrunn (approved):

hbrunn:
looks like we need another rebase

luc-demeyer:
PR https://github.com/OCA/l10n-netherlands/pull/111 is the one causing this rebase, but that one must be fixed first (cf. my comments in that PR).
I'll rebase once the solution for this issue is agreed upon.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -96,6 +90,7 @@ def check_periods(self):
     @api.multi
     def button_generate(self):
         self.date_generated = fields.Datetime.now(self)
+        self._get_data()
```
StefanRijnhart: It seems odd that you store the data on the Model object. Why don't you return the data structures in the call to `_get_data()` and include the data structures in the `values` dictionary in the call to `render`?

luc-demeyer: This approach works and is performant (but could create problems in inherited modules).
Your suggestion to pass the date via the values dict would solve that.
Please feel free to make a PR on top of this PR and I'll review/test the code and merge so that we can proceed with merging this functional enhancement.

StefanRijnhart: Luc, I'm reviewing your code here so I am a bit surprised that you ask me to propose a PR for a simple change like this. It seems that you think the suggestion is an improvement, but I can imagine that you think the risk is too theoretical (I believe it is two simultaneous exports on the same Odoo instance that will cause the problem) and not worth the hassle so then we can just leave it to the other reviewers to support the suggestion or not.

luc-demeyer: stefan,
I would need to investigate if this code could fail with two simultaneous exports on the same instance but I agree that we better adapt the code. I'll put in on my todo list and try to free a bit of time before the end of next week (but I don't mind if you have time to do it earlier).

StefanRijnhart: As a committer on this repository, I can push to your branch directly. If you don't mind that, I can have a look. It is more convenient because I can check travis results directly (and fix them).

luc-demeyer: No problem, go ahead.

StefanRijnhart: Done. I also implemented @hbrunn's suggestion to have the additional template inherit from the original one. You might want to check that one.

hbrunn:
@astirpe @StefanRijnhart do you agree to merge?

StefanRijnhart:
I'm a little surprised that no test was added to reproduce the problem (irrelevant parters/accounts/journals in the export) and ensure that no regression is introduced in the future but I won't block a merge for that now.

StefanRijnhart (approved):

StefanRijnhart (changes requested):
Actually, a more important test is missing: no 'full' export is created during testing, only a 'default' test. That should be easy to fix.

luc-demeyer:
stefan,

Thanks for the work done.
I reviewed the code and tested it.  Imho it's ok to merge.
I have also added one more basic unit test and agree with your remark that the set of unit tests is far from complete (but I also agree that we shouldn't block the merge because of this).

hbrunn:
well, my rule (actually, that's my reading of the [conventions](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#it-makes-sense-to-be-picky-in-the-following-cases), third bullet, but a lot of community members seem to read this differently) is simple: A PR cannot decrease code coverage. That's not the case here, so I wouldn't have blocked, but of course it's nicer with a test, thanks

StefanRijnhart (approved):
Thanks for that extra test!

PR #120, opened 2017-10-05
[10.0] [IMP] l10n nl partner salutation gender

NL66278:
- Resolve incompatibilities with partner_contact_gender.
- Improved handling of company partners.
- Improved OCA standard compliance.
- Improved tests.
- Recognize difference between unknown gender (will be False/NULL) and other.


NL66278:
Runbot errors are due to problems with runbot - as usual

l10n_nl_partner_salutation/models/res_partner.py
```diff
@@ -3,7 +3,21 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import api, models, fields
 
-GENDERS = [('male', 'Male'), ('female', 'Female'), ('unknown', 'Unknown')]
+
+DEFAULT_PREFIX = {
+    'shortcut': {
+        'male': 'De heer',
```
hbrunn, line 9: what i disliked with the 8 version is that we'll never be able to address non dutch speaking people with the dutch regime. Better put this in a function of its own to return the dictionary, then you can wrap the values in `_()`. And you also enable other modules to hook in there and do what they think is the good default.

l10n_nl_partner_salutation/models/res_partner.py
```diff
@@ -19,15 +33,21 @@ def on_change_use_manual_salutations(self):
...
-            "${' ' if infix else ''}"
-            "${lastname or ''}"
+    @api.multi
+    def get_salutation_name(self):
+        """Return the name formatted for use in salutation."""
+        self.ensure_one()
+        # Use API from partner_firstname
+        # (actually method as it has been overridden in partner_name_dutch)
+        return self.with_context(
+            name_format="${firstname or initials or ''}"
```
hbrunn: the idea for having `get_salutation_name_format` in its own function was modularity, so other modules can change that format as they please. Any reason to destroy this possibility?

hbrunn (changes requested):

hbrunn:
@NL66278 can you fix [travis](https://travis-ci.org/OCA/l10n-netherlands/jobs/285796831)? then we'll be able to merge this

hbrunn (approved):

PR #122, opened 2017-11-25
[11.0][MIG] l10n_nl_tax_invoice_basis

astirpe:
Port of l10n_nl_tax_invoice_basis to v11


astirpe:
Oops!
Sorry Runbot:

```
createdb: database creation failed: ERROR:  could not write to file "base/24484/12007": No space left on device
```

l10n_nl_tax_invoice_basis/models/account_tax.py
```diff
@@ -0,0 +1,57 @@
...
+                res = self._get_invoice_basis_domain(domain_params)
+        return res
+
+    def _get_invoice_basis_domain(self, domain_params):
+        tax_date_domain = self._get_tax_date_domain(domain_params)
+        date_domain = self._get_accounting_date_domain(domain_params)
+        return [
+            ('company_id', '=', domain_params['company_id']),
+            '|',
+        ] + tax_date_domain + date_domain
```
hbrunn, line 41: I like using `expression.AND` / `expression.OR` in such cases, but I won't block on this

l10n_nl_tax_invoice_basis/models/account_tax.py
```diff
@@ -0,0 +1,57 @@
...
+class AccountTax(models.Model):
+    _inherit = 'account.tax'
+
+    def get_move_line_partial_domain(self, from_date, to_date, company_id):
+        res = super(AccountTax, self).get_move_line_partial_domain(
+            from_date,
+            to_date,
+            company_id
+        )
+        if not self._context.get('skip_invoice_basis_domain'):
```
hbrunn: shouldn't you check here if your flag on the company is set too?

hbrunn: and for stopping condition, I always recommend

```
if $stopping_condition:
    return $result
```

because then you save a level of indentation, and it's clearer what's going on imho

hbrunn (changes requested):

l10n_nl_tax_invoice_basis/models/account_tax.py
hbrunn: so https://github.com/OCA/l10n-netherlands/pull/122#discussion_r156057093 is my only real request for change, the rest take or leave as you prefer

astirpe: With latest commit I just return if the stopping_condition is true.

The idea to keep two separate checks, one for the 'skip_invoice_basis_domain' in the context and the other for the country, is that:
if 'skip_invoice_basis_domain' is defined in the context, I skip the next check: the browse of the company and the subsequent check on its country. The intention is to save a prefetch call by the ORM, in case the context already triggered the stop condition.

melroy89 (approved):

melroy89:
@pedrobaeza Please merge.

pedrobaeza:
Sorry, but this is as localization repository and I shouldn't do that task as general community coordinator, being very country specific and not being able to have an opinion on this. There are several PSC members that can do that if appropriated.

l10n_nl_tax_invoice_basis/models/account_tax.py
melroy89: @astirpe Should we merge it or do you want to change the line with the remark?

astirpe: I would prefer to keep this part of code as it is.

hbrunn:
@astirpe can you fix the conflict? then i'll merge

hbrunn (approved):

hbrunn:
PS: while being at it, please squash your v11 commits

astirpe:
@hbrunn I probably made a mistake: I used the Github interface to resolve the conflicts and now I got a merge with branch 11.0. This causes a lot of unwanted commits in my commits history, see https://github.com/onesteinbv/l10n-netherlands/commits/11_mig_l10n_nl_tax_invoice_basis. Is there a way to revert the merge?

hbrunn:
hmmm, I never really used github's edit functions, save for some typos. What I'd do is to squash the commits locally and force push

astirpe:
It seems it's ok now, may I ask you to double check?

melroy89:
LGTM, lets merge it?

Issue #125, opened 2017-12-01
 l10n_nl_tax_statement -  4a & 4b turnover negative


CasVissers-360ERP:
Hi,

When creating a tax statement the turnover in 4a & 4b are shown as a negative amount. I think this should be positive, surely because the tax in 4a&4b is shown as a positive number.

Cas

yung-wang:
Made a test in V10 with and agree. It is best to make the turnover amount positive. Can we make the display of the turnover amount into an absolute?

astirpe:
Sure, I will fix it next week. Thank you!

astirpe:
5a is the sum of 1a through 4b.
Is the actual calculation of 5a correct?

In case I display a positive amount of 4a and 4b, shall I also change the calculation of 5a?

CasVissers-360ERP:
The calculation of 5a is correct. Its just the case of displaying the amounts as positive.

astirpe:
@CasVissers can you check the PR #127 ?

CasVissers-360ERP:
@astirpe 
Looks good thus far. 
One thing I missed in my ticket was 2a, this fix should apply to 2a as well.

astirpe:
@CasVissers shouln't be applied for all the values, then? From 1a to 4b, if the omzet has a different sign than btw, invert the omzet sign. What do you think?

CasVissers-360ERP:
@astirpe 
You are right, I agree that's the best solution.

astirpe:
Done, can you check?

CasVissers-360ERP:
Done, issue closed.

PR #127, opened 2017-12-04
[10.0][FIX] Omzet 4a-4b: check and invert the sign

astirpe:
Fixes https://github.com/OCA/l10n-netherlands/issues/125


yung-wang (approved):
I checked with a few invoices. The amount is now shown as a positive. I approve it.

CasVissers-360ERP (approved):
Tested and approved.

PR #128, opened 2017-12-11
[10.0] l10n_nl_partner_name improvements

hbrunn:
this started with wanting to have the form look like

screenshot_20171211_124726

the rest are fixes I came across on the way. I'm of a mind to propose a similar patch for `partner_firstname`, but that's slightly more tricky because there, we've got to cope with different name orders.

Another thing I've in mind (but nobody paying for it) is to make infix and prefix configurable, possibly on group level. Then users can decide if the want to use only infixes, only prefixes or both. The forms then can be adapted with appropriate group specific views

depends on https://github.com/OCA/server-tools/pull/1090


hbrunn:
PS: Contrary to the screenshot, I swapped first name and initials after talking to my colleagues

NL66278:
schermafdruk van 2018-01-27 00-06-01

Maybe it is our particular installation (and we should definitely move the gender field). Still we get all the name fields except firstname very small and confined to less then half of the screen.

hbrunn:
@NL66278 looks like there's still a [button box](https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/less/form_view_layout_extra.less#L48) or an [oe_left or oe_right](https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/less/form_view_layout.less#L10) active on your form

lfreeke:
@hbrunn The initials and infix field are very small. You hardly can see what you type. Should we add some more width to the fields?

NL66278:
@hbrunn Could you please rebase this one? We have a merge conflict now.

hbrunn:
@NL66278 done. Please add your formal review so that we're a step closer to merge this in order to avoid similar problems in the future.
I assumed a 'normal' initial and infix to be usually two characters wide, what would be better values in order to fix @lfreeke's comment?

RoelAdriaans:
Maybe bump the version, because this pull requests changes views, that can break inherited views?

NL66278 (approved):
:+1:  Code review and test

NL66278:
@StefanRijnhart This no longer needs fixing (has been done), but I cannot remove the label.

CasVissers-360ERP:
@StefanRijnhart LGTM

Only possible improvement can be to increase infix to 7 characters (van der)

CasVissers-360ERP (approved):
Functional test (in v11)

StefanRijnhart (approved):

CasVissers-360ERP:
@hbrunn Only possible improvement can be to increase infix to 7 characters (van der) 
Can you decide on this?

PR #129, opened 2017-12-12
[10.0] l10n_nl_xaf_auditfile_export error with zero amounts

RoelAdriaans:
If the amount is zero, query returns nothing,
and generation will fail with error:

:31681:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://www.auditfiles.nl/XAF/3.2}totalDebit': '' is not a valid value of the atomic type '{http://www.auditfiles.nl/XAF/3.2}Amount2decimals'.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -225,7 +225,7 @@ class XafAuditfileExport(models.Model):
             'and date <= \'' + self.date_end + '\' '
             'and (company_id=%s or company_id is null)',
             (self.company_id.id, ))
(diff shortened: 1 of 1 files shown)


hbrunn:
please cherry pick the bugfix from 8: https://github.com/OCA/l10n-netherlands/pull/121

RoelAdriaans:
@hbrunn Didn't see that, same issue :)

I cannot do a git cherry-pick, because there are too much changes between the versions.
Any idea what do to?

hbrunn:
now seeing your second PR I'm quite confident that it fixes this issue (in a better way than I did) too, so my suggestion would be to do nothing and close this PR if that's true

RoelAdriaans:
Also fixed by #130 , closing this one.

PR #131, opened 2017-12-13
WIP 8.0 l10n nl hr employee name

gfcapalbo:


l10n_nl_hr_employee_name/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
hbrunn, line 20: use short license headers (and update your templates)

l10n_nl_hr_employee_name/__openerp__.py
```diff
@@ -0,0 +1,30 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Dutch Employee names',
+    'version': '8.0.1.0.0',
+    'author': 'Therp BV, Odoo Community Association (OCA)',
+    'category': 'Contact management',
+    'depends': [
+        'hr',
```
hbrunn, line 27: no `l10n_nl_partner_name`?

l10n_nl_hr_employee_name/model/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from . import hr_employee
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
hbrunn, line 24: no vim codelines

l10n_nl_hr_employee_name/model/hr_employee.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class HrEmployee(models.Model):
+    """Extend hr.employee with extra fields for Dutch names. Reusing the
+    firstname, lastname, initials and infix from  partner_firstname and
+    l10n_nl_partner_name, simpler and more consistent than repeating the logic
+    """
+    _inherit = 'hr.employee'
+
+    firstname = fields.Char(string='firstname')
```
hbrunn, line 33: no string in this case (used default field name)

l10n_nl_hr_employee_name/model/hr_employee.py
```diff
@@ -0,0 +1,65 @@
...
+    lastname = fields.Char(string='lastname')
+    initials = fields.Char(string='initials')
+    infix = fields.Char(string='infix')
+    name_informal = fields.Char(string='nickname')
+    name_formal = fields.Char(
+        string='Name', compute='_compute_display_name_formal', store=True)
+    name_informal = fields.Char(
```
hbrunn: those are specific to our customer, remove here

l10n_nl_hr_employee_name/model/hr_employee.py
```diff
@@ -0,0 +1,65 @@
...
+    """Extend hr.employee with extra fields for Dutch names. Reusing the
+    firstname, lastname, initials and infix from  partner_firstname and
+    l10n_nl_partner_name, simpler and more consistent than repeating the logic
+    """
+    _inherit = 'hr.employee'
+
+    firstname = fields.Char(string='firstname')
+    lastname = fields.Char(string='lastname')
+    initials = fields.Char(string='initials')
+    infix = fields.Char(string='infix')
```
hbrunn, line 36: add `nickname`

l10n_nl_hr_employee_name/scripts/guess_names.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/python
```
hbrunn, line 1: I think while being at it, we should make this a little wizard you can call on partners. nothing fancy, just one shoot do it, maybe a dialog where stuff like the list of strings to be treated as infix and the regex can be changed, then this could be useful for other languages too.

or integrate this into https://github.com/OCA/partner-contact/blob/8.0/partner_firstname/models/res_partner.py#L146, but that might become very tricky for the general case. If it turns out too tricky, just do the wizard on `active_ids`

l10n_nl_hr_employee_name/tests/test_l10n_nl_hr_employee_name.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
+
+
+class TestL10nNlPartnerName(TransactionCase):
```
hbrunn, line 7: from the looks of it, you should be able to import the test case classes from https://github.com/OCA/partner-contact/blob/8.0/partner_firstname/tests, override some strategic functions where you create employee records instead of partners, and get all the test code for free thanks to late binding (only works if you follow my advice above to recreate the `protocol` of partner_firstname by providing functions with the same name.

remember that unittest autodetects tests classes by class name, so you must alias the classes you import or just import the module containing it.

l10n_nl_hr_employee_name/view/hr_employee.xml
```diff
@@ -0,0 +1,32 @@
...
+      <field name="name">hr.employee.simplified.form.firstname</field>
+      <field name="model">hr.employee</field>
+      <field
+          name="inherit_id"
+          ref="hr.view_employee_form"
+      />
+      <field name="arch" type="xml">
+          <data>
+              <label for="name" position="replace"/>
+              <xpath expr="//h1[field[@name='name']]" position="replace">
```
hbrunn, line 13: I think we should do something similar to https://github.com/OCA/l10n-netherlands/pull/128#issue-280988907 here

hbrunn (changes requested):
what we both missed is to check if https://github.com/OCA/hr/tree/8.0/hr_employee_firstname exists. it does, so we should inherit from it

hbrunn:
see https://github.com/OCA/l10n-netherlands/pull/132#issuecomment-358000791

PR #134, opened 2017-12-21
[FIX] Cast config param to int

daramousk:


hbrunn (approved):

astirpe (approved):
This fix is also needed for 9.0, 10.0 and 11.0

hbrunn:
absolutely!

PR #136, opened 2017-12-29
[8.0][IMP] l10n_nl_xaf_auditfile_export. Speedup with SQL and direct IO.

NL66278:
Creating an auditfile for a not even to large dataset was excruciatingly slow or did not complete at all.

Reworking the xml creation allows for fast export of data (11 minutes for a 386 Mb. export file on a small system with limited resources).

As base64 encoding to create an attachment also caused Out of Memory Orders, the created audit files are now stored directly in the filesystem. A special controller has been created to serve these auditfiles.

Still need proper argument handling for auditfile.py when called as a script.


hbrunn:
I don't think we can merge this, because it will break stuff like https://github.com/OCA/l10n-netherlands/pull/119 - we lose the simple overridability here. Given that QWeb is rewritten in 10 such that it basically does what you propose here, we'll be back to QWeb there.
An extra module however that people can install if the standard version fails would be great.

NL66278:
@hbrunn seems to me that the changes in #119 can be easily incorporated in this one.

I doubt very much whether the qweb compile will do enough to reduce all of the orm overhead in creating the xml for large (actually only medium) sized databases, but we will see that when we move to 10.0.

l10n_nl_xaf_auditfile_export/controllers/main.py
```diff
@@ -0,0 +1,22 @@
...
+from openerp import http
+from openerp.http import request
+from openerp.addons.web.controllers.main import content_disposition
+
+from ..models.xaf_auditfile_export import get_auditfile_path
+
+
+class Main(http.Controller):
+
+    @http.route('/file_cabinet/auditfile', type='http', auth='user')
```
StefanRijnhart, line 13: ~~This strikes me as idiosyncratic. I mean, there is no technical reason not to store this as an attachment any more, right?~~
I see that you had a memory error with the base64. I would still be in favour of another solution.

Seems like the auditfile can be downloaded by any user if they know the url?

l10n_nl_xaf_auditfile_export/auditfile.py
```diff
@@ -0,0 +1,641 @@
...
+    writeln_element_close(outfile, 'header', indent=2)
+
+
+def patch_country_code(country_code):
+    """Not all countrycodes are already in the xsd schema.
+
+    JE - Jersey
+    ME - Montenegro
+    """
+    if country_code in ['JE', 'ME']:
```
StefanRijnhart, line 322: That is a little sneaky, hiding an unrelated change like this in a big PR. Did you change any thing else functionally?

StefanRijnhart: I created https://github.com/OCA/l10n-netherlands/pull/166/files earlier today which addresses the same problem.

StefanRijnhart:
I got hit by the memory error as well, and I am proposing https://github.com/OCA/l10n-netherlands/pull/167

hbrunn:
superseded by #167 167

daramousk:
This implementation seems to be orders of magnitude faster than #167. I export a yearly period of data and end up with a 600mb file. And this all happens under 10 minutes. On #167 it takes at least an hour before it crashes. Maybe all the changes should be incorporated into this branch...

StefanRijnhart:
Can you check what the bottleneck is in your case? Is it fetching the data, generating the XML or validating the XML? You can determine what part of this PR to pick when we know that.

PR #138, opened 2018-01-08
Incorrect link to community website

melroy89:
Fix link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@ Odoo Dutch localization project
 ===============================
 
 This repository includes all the Odoo 11.0 modules developed to fit the specific needs of the Netherlands.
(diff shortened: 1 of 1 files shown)


hbrunn:
thanks for this, but the OCA website is still a bit in flux after the update to 10 a few weeks ago, so I expect the project links to work again in a while. SSL links are a good choice then anyways. I suggest to keep this open for a while to see how the OCA website develops, we want to link to the nl-specific page because it contains info like the mission statement, the OCA representatives etc

melroy89:
OK I see, thank you for thinking along.

melroy89:
Let's close it, and see if they fix the website anytime soon.

PR #130, opened 2017-12-12
[IMP] Round floats to two decimals

RoelAdriaans:
We use a database with multi-currency, and because Odoo's habbit to use floats, we also have floats in the database. 

Report does no rounding, giving us the wonderfull errors:
<pre>
 :140197:0:ERROR:SCHEMASV:SCHEMAV_CVC_FRACTIONDIGITS_VALID: Element '{http://www.auditfiles.nl/XAF/3.2}amnt': [facet 'fractionDigits'] The value '1183.03194186' has more fractional digits than are allowed ('2').

:140197:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://www.auditfiles.nl/XAF/3.2}amnt': '1183.03194186' is not a valid value of the atomic type '{http://www.auditfiles.nl/XAF/3.2}Amount2decimals'.

:140209:0:ERROR:SCHEMASV:SCHEMAV_CVC_FRACTIONDIGITS_VALID: Element '{http://www.auditfiles.nl/XAF/3.2}amnt': [facet 'fractionDigits'] The value '1183.03194186' has more fractional digits than are allowed ('2').

:140209:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://www.auditfiles.nl/XAF/3.2}amnt': '1183.03194186' is not a valid value of the atomic type '{http://www.auditfiles.nl/XAF/3.2}Amount2decimals'.

:143824:0:ERROR:SCHEMASV:SCHEMAV_CVC_FRACTIONDIGITS_VALID: Element '{http://www.auditfiles.nl/XAF/3.2}amnt': [facet 'fractionDigits'] The value '1183.03194186' has more fractional digits than are allowed ('2').

:143824:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://www.auditfiles.nl/XAF/3.2}amnt': '1183.03194186' is not a valid value of the atomic type '{http://www.auditfiles.nl/XAF/3.2}Amount2decimals'.

:143836:0:ERROR:SCHEMASV:SCHEMAV_CVC_FRACTIONDIGITS_VALID: Element '{http://www.auditfiles.nl/XAF/3.2}amnt': [facet 'fractionDigits'] The value '1183.03194186' has more fractional digits than are allowed ('2').

[...]

PR #146, opened 2018-01-31
XmlAuditfileFinancieel syntax fix

CGenie:
Some leftover + from git merging was in the file which resulted in
validation errors.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/views/templates.xml
+++ b/l10n_nl_xaf_auditfile_export/views/templates.xml
@@ -148,7 +148,7 @@
                         <docRef><t t-esc="l.ref" t-esc-options='{"widget": "auditfile.string999"}' /></docRef>
                         <effDate><t t-esc="l.date" /></effDate>
                         <desc><t t-esc="l.name" /></desc>
-                        <amnt><t t-esc="l.credit or l.debit" /></amnt>
+                        <amnt><t t-esc="abs(l.balance)" /></amnt>
                         <amntTp><t t-esc="'C' if l.credit else 'D'" /></amntTp>
                         <recRef t-if="l.full_reconcile_id"><t t-esc="l.full_reconcile_id.name" /></recRef>
                         <custSupID t-if="l.partner_id"><t t-esc="l.partner_id.id" /></custSupID>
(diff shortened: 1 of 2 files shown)


astirpe (approved):

l10n_nl_xaf_auditfile_export/views/templates.xml
```diff
@@ -148,7 +148,7 @@
                         <docRef><t t-esc="l.ref" t-esc-options='{"widget": "auditfile.string999"}' /></docRef>
                         <effDate><t t-esc="l.date" /></effDate>
                         <desc><t t-esc="l.name" /></desc>
-                        <amnt><t t-esc="l.credit or l.debit" /></amnt>
+                        <amnt><t t-esc="round(l.credit or l.debit, 2)" /></amnt>
```
astirpe: I think you don't need to round here. The risk is that the rounding will generate an amount different than `l.balance`. See https://github.com/OCA/l10n-netherlands/pull/130#issuecomment-353287014

astirpe (changes requested):

l10n_nl_xaf_auditfile_export/views/templates.xml
CGenie: Well we had this error:
```
:507419:0:ERROR:SCHEMASV:SCHEMAV_CVC_FRACTIONDIGITS_VALID: Element '{http://www.auditfiles.nl/XAF/3.2}amnt': [facet 'fractionDigits'] The value '0.00999999999999' has more fractional digits than are allowed ('2').
```

astirpe: Isn't abs(l.balance) fixing that issue as well?

CGenie: Yes, it works, thanks.

astirpe (approved):

hbrunn:
please squash your commits to one about the country, one about the rounding

hbrunn (approved):

CGenie:
There need to be 3 commits as they fix 3 different things: syntax error in xsd, JE country and rounding error.

hbrunn:
ah right, the -+ line fooled me. Will merge after green runbot

Issue #112, opened 2017-09-02
CBS export Bugs


ploegvde:
When testing by our customer we have noted a few bugs. We will create PR soon for the fixes:

1. First line, phone number. If phone number of company setting is with country code =31, it has to be replaced by a 0
2. First line pos 22: This digit has to represents the month that it concerns. Now it is wrongly the month of creation of the file. 
3. First line: Change line ending  LF  > CRLF
4. Detail lines pos 100. This has to be a invoice line number increment from 00 until 99.
5. Detail lines: Change line ending  LF  > CRLF
6. detail lines: If amount is 0, then calculate the line value as Cost price x qty
7. Footer line: Change line ending  LF  > CRLF

PR #152, opened 2018-02-19
[11.0][MIG] l10n_nl_cbs_export

astirpe:
Porting `l10n_nl_cbs_export` to V11.
Includes fixes for issue https://github.com/OCA/l10n-netherlands/issues/112


StefanRijnhart (changes requested):
Diff looks good! Would it be possible to attribute the work of the original author of the fixes for #112 with a separate commit using --author?

astirpe:
@StefanRijnhart done. Thanks!

l10n_nl_cbs_export/models/cbs_export_file.py
```diff
@@ -0,0 +1,279 @@
...
+                    '-', '').zfill(10)
+            else:
+                value += str(int(invoice_line.price_subtotal)).replace(
+                    '-', '').zfill(10)
+            value += str('+') + str('0000000000').zfill(10)
+
+            if len(str(invoice_line.invoice_id.number or '')) < 8:
+                invoice_value = \
+                    str(invoice_line.invoice_id.number) + \
+                    str(invoice_line_number).zfill(2)
```
StefanRijnhart, line 260: You could use `(line_counter -1) % 100` so that you don't have to keep two counters. I find having two counters slightly confusing. But it's a matter of taste, of course (non blocking for merge).

StefanRijnhart (approved):
Thanks for going through the trouble of reorganizing those commits!

CasVissers-360ERP (approved):
Basic functional test done.

l10n_nl_cbs_export/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2017 Odoo Experts (<https://www.odooexperts.nl>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "CBS Export Intrahandel Sale",
+    "version": "11.0.1.0.0",
+    "author": "Odoo Experts, Onestein, Odoo Community Association (OCA)",
+    "website": "https://www.odooexperts.nl",
```
tarteo: Can you change the website to https://github.com/OCA/l10n-netherlands

l10n_nl_cbs_export/models/cbs_export_file.py
```diff
@@ -0,0 +1,279 @@
...
+                str(datetime.strptime(
+                    invoice_line.invoice_id.date_invoice, '%Y-%m-%d'
+                ).strftime("%Y%m") or '').ljust(6) + \
+                str('7') + \
+                str(invoice_line.company_id.vat or ''
+                    ).replace(' ', '')[2:].ljust(12) + \
+                str(line_counter).zfill(5) + str(" " * 3) + \
+                str(invoice_line.invoice_id.partner_id.country_id.code or ''
+                    ).ljust(3) + \
+                str('3') + \
```
tarteo, line 233: Why use str() here? It's not really clear by reading what is happening here.

tarteo (changes requested):

l10n_nl_cbs_export/models/cbs_export_file.py
astirpe: @tarteo I agree, the code should be refactored.

As explained [here](https://github.com/OCA/l10n-netherlands/pull/84#issuecomment-306746876):

> ... the fixed values we use are for most common situations correct

and in the readme:

> Field Transport "Vervoerwijze" is set hardcoded on 3. This is the most common used value.

I would like to do refactor and complete the module according to the Roadmap, but in a separate PR. What do you think?

tarteo: In a seperate PR is OK for me.

tarteo (approved):

astirpe:
Thanks!

PR #154, opened 2018-03-06
[FIX] l10n_nl_xaf_auditfile_export: fix rounding issues

Yenthe666:
Before this fix you will get both an error on the `totalCredit` and `totalDebit` when the result of one of those two has a bigger decimal precision than 2. For an example customer we would get this traceback:
```
:18296:0:ERROR:SCHEMASV:SCHEMAV_CVC_FRACTIONDIGITS_VALID: Element '{http://www.auditfiles.nl/XAF/3.2}totalCredit': [facet 'fractionDigits'] The value '1207618.6192' has more fractional digits than are allowed ('2').

:18296:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://www.auditfiles.nl/XAF/3.2}totalCredit': '1207618.6192' is not a valid value of the atomic type '{http://www.auditfiles.nl/XAF/3.2}Amount2decimals'. 
```

By rounding the returned SQL values we're always sure the precision is 2 and not something like 4.
This fixes two issues at once. I'm just not sure if we should do it with a round and precision 2 on the return or with a round in the SQL query.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -222,7 +222,7 @@ class XafAuditfileExport(models.Model):
             'and date <= %s '
             'and (company_id=%s or company_id is null)',
             (self.date_start, self.date_end, self.company_id.id, ))
-        return self.env.cr.fetchall()[0][0]
+        return round(self.env.cr.fetchall()[0][0], 2)
 
     @api.multi
     def get_move_line_total_credit(self):
@@ -233,7 +233,7 @@ class XafAuditfileExport(models.Model):
             'and date <= %s '
             'and (company_id=%s or company_id is null)',
             (self.date_start, self.date_end, self.company_id.id, ))
-        return self.env.cr.fetchall()[0][0]
+        return round(self.env.cr.fetchall()[0][0], 2)
 
     @api.multi
     def get_journals(self):


hbrunn:
if you want to change anything, using [float_round](https://github.com/OCA/OCB/blob/11.0/odoo/tools/float_utils.py#L37) might be a good idea for consistency. Even though I can't think of a case where this will behave differently than `round`. I'd merge this as it is tbh

Yenthe666:
@hbrunn hmm yes I was not sure there either, I couldn't think of any other cases either. By the way, this issue should be identical in 9.0 and 10.0 now that I think of it. Do you guys forward port like Odoo too? Or do you want me to make three PR's?

hbrunn:
I would *like* you to do three PRs, but there's no way I can *ask* you to do this I think.

Forward/backward porting in the OCA relies on somebody needing whatever is done in another version, so no, there's no central organization that keeps versions in sync.

Yenthe666:
I'm more then happy to fix it in all three versions at once, so it is handled and identical in all versions. I'll wait for now to be sure that this will get accepted and merged in. If this is the case I'll make PR's for 9.0 and 10.0 too.
CC @astirpe and @RoelAdriaans-B-informed by the way.

StefanRijnhart:
`self.company_id.currency_id.round()` which uses float_round with the currency's precision. But then, what are the chances that the currency is going to be anything other than EUR....

StefanRijnhart (approved):

Yenthe666:
Yep, agreed. So what more do I need to get this merged guys? Or should I still change it to the currency_id round?

hbrunn:
I just wait for somebody to scream `merge this'

Yenthe666:
Hmm. Merge this! There, I did it haha. ;)

PR #155, opened 2018-03-06
[FIX] CSS Namespace conflict with account_financial_report_qweb

StefanRijnhart:
The CSS in this module seems borrowed from https://github.com/OCA/account-financial-reporting/blob/10.0/account_financial_report_qweb/static/src/css/report.css. The styles have diverged by now, so that when both modules are installed the layout is messed up.

vat


astirpe (approved):
Thank you @StefanRijnhart

PR #158, opened 2018-03-07
[FIX] force periods in tests

hbrunn:
the wizard's `default_get` selects the previous fiscal year if it finds any. Now there are a couple of other modules that create fiscal years in the past for testing purposes, which will make the current tests break because then we select this earlier fiscal year, but the demo move lines are created for the current fiscal year. With the filtering introduced by #119, we'll end up without accounts, rendering the xaf invalid and the tests failing.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
@@ -8,13 +8,25 @@ from openerp.tests.common import TransactionCase
 
 class TestL10nNlXafAuditfileExport(TransactionCase):
 
+    def setUp(self):
+        super(TestL10nNlXafAuditfileExport, self).setUp()
+        self.fiscalyear = self.env['account.fiscalyear'].browse(
+            self.env['account.fiscalyear'].find()
+        )
+        self.export = self.env['xaf.auditfile.export'].create({})
+        # depending on other modules installed, we get some undefined
+        # fiscal year via defaults, force this to the current year
+        # in order to be sure we get account's demo data
+        self.export.write({
+            'period_start': self.fiscalyear.period_ids[0].id,
+            'period_end': self.fiscalyear.period_ids[-1].id,
+        })
+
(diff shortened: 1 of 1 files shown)


astirpe (approved):

l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
```diff
@@ -8,13 +8,25 @@
 
 class TestL10nNlXafAuditfileExport(TransactionCase):
 
+    def setUp(self):
+        super(TestL10nNlXafAuditfileExport, self).setUp()
+        self.fiscalyear = self.env['account.fiscalyear'].browse(
+            self.env['account.fiscalyear'].find(exception=False) or []
```
StefanRijnhart: You seem to cover the case in which there is no current fiscal year. Does that not break the tests below?

hbrunn: no. What we need to do here is to take the current fiscal year because some other module creates an old fiscal year with no invoices

StefanRijnhart: So the empty list is a valid fallback? I mean, does that not mean that there is no current fiscal year which would break the generation of the auditfile for this non existing fiscal year? Intuitively I would just call find without disabling the exception or the empty fallback value.

hbrunn: ah, that's what you mean. No, that's just a habit

StefanRijnhart: Do you want to leave it in? As you can see, it confuses the simpler minds like me.

hbrunn: I changed the whole thing to also not decrease coverage

StefanRijnhart (approved):

PR #167, opened 2018-05-14
auditfile/[IMP] Memory footprint, speed and disk space

StefanRijnhart:
Generating and storing a 70MB auditfile takes 8 minutes for qweb, but for
lxml.etree to reconstruct in order to apply a namespace takes 3 hours and 1GB
of extra memory.

Further optimization is achieved by sticking to the Odoo prefetch size when
chunking record sets.

As XML compresses very well, the auditfile is now stored in the Odoo database
in compressed format for easier handling and saving disk space.


StefanRijnhart:
@hbrunn  I'm very sorry. My way of applying the namespace is a travesty but I think the performance increase is worth it.

hbrunn:
I'm a bit embarrassed to read this code. But well, that's part of the game, we hate our own code a few years after the fact. The huge memory footprint stems from my bad understanding of lxml back then. We can just construct a generator like the one I use there, but feed that directly to the validator, starting with our handcrafted node and then feed the children from a generator that reads the qweb doc from a generator. then we should have a constant memory footprint without string manipulations.

StefanRijnhart:
Dont be silly, you saved everyone's hide by creating this module in the first place ;-)

I'll have a look at what your suggestion means for the footprint.

StefanRijnhart:
@hbrunn I thought long and hard about what you are suggesting and I don't get it. The problem with iterparse is that it provides events, not nodes. The root node is only available after the whole structure has been read, so we cannot free up memory iteratively. Or are you suggesting something more low level?

hbrunn:
no, iterparse + validation during parsing should work: http://lxml.de/validation.html#validation-at-parse-time

StefanRijnhart:
My question was more about how to create a generator with a namespace from a generator if that generator gives me events, not nodes. It seems I need everyting in memory before I can iterate over the nodes, which makes me think it defeats the purpose. Sorry, still at a complete loss.

hbrunn:
the first element of the generator is the node. This you should be able to manipulate to add the namespace to the root element, and adapt the elements the way we do it there already. This assumes that validation only happens after our manipulation, but it's my understanding that's the case

StefanRijnhart:
Thanks for your help! I got this far
```
        first = True
        for action, element in iterparse:
            if first:
                root = element
                while root.getparent():
                    root = root.getparent()
                root.nsmap = nsmap
                first = False
            element.tag = '{%s}%s' % (default_namespace, element.tag)
```
but then *** AttributeError: attribute 'nsmap' of 'lxml.etree._Element' objects is not writable

I also cannot set individual keys in root.nsmap. That fails silently. Any other idea how to manipulate the root's namespace?

hbrunn:
I fiddled a bit with this while waiting for a build, I came up with this:

```python
from lxml.etree import iterparse, XMLSchema
from StringIO import StringIO

xml = StringIO('<hello><world/></hello>')
schema = XMLSchema(file=StringIO(
    '<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema">'
    '<xsd:element name="hello">'
    '<xsd:complexType><xsd:sequence>'
    '<xsd:element name="world" />'
    '</xsd:sequence></xsd:complexType>'
    '</xsd:element>'
    '</xsd:schema>'
))
parser = iterparse(xml, events=('end',), schema=schema)
for event, element in parser:
    element.clear()
```

can you check how this behaves with your huge file?

StefanRijnhart:
Are we talking about the same thing here? I believe the challenge at hand is still your desire to add the namespace other than using a string replace (which works fine IMHO). I think your snippet does not deal with namespaces, or is it the heat?

If I run the unittest of the module with this I get 
`lxml.etree.XMLSyntaxError: Element 'auditfile': No matching global declaration available for the validation root.`

It's this code: 
```
        schema = etree.XMLSchema(etree.parse(
            file(
                modules.get_module_resource(
                    'l10n_nl_xaf_auditfile_export', 'data',
                    'XmlAuditfileFinancieel3.2.xsd'))))

        iterparse = etree.iterparse(
            StringIO(xml), events=('end',), remove_blank_text=True, remove_comments=True,
            schema=schema)

	for action, element in iterparse:
            element.clear()
```

hbrunn:
not really any more ;-) I was focusing on the validate-stuff-without-loading-the-whole-file part, but dropped the namespace part. This you should be able to add in the start event the way the original code does it, and then the nodes are namespaced and the whole thing should validate

StefanRijnhart:
Not sure what you mean. The original code modified the tag on the elements. Is that the same as 'add in the start event'? Do I keep the filter on 'end' events? Or again, can't we just keep the string replacement?

hbrunn:
let's keep the string replace

hbrunn (approved):

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -24,7 +37,7 @@ class XafAuditfileExport(models.Model):
 
     @api.depends('name')
     def _auditfile_name_get(self):
-        self.auditfile_name = '%s.xaf' % self.name
+        self.auditfile_name = '%s.xaf.zip' % self.name
```
astirpe: May I suggest to make field `auditfile_name` as stored? We need to preserve the original name in the old generated auditfiles.

Moreover, the calculation of the file name could be deducted from the file type. What I mean is this: 
https://github.com/onesteinbv/l10n-netherlands/blob/11_imp_l10n_nl_xaf_auditfile_export2/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py#L42-L50

astirpe (changes requested):

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
StefanRijnhart: Looks like you have it all set up in your branch. Care to make a PR into mine?

astirpe: Here it is https://github.com/StefanRijnhart/l10n-netherlands/pull/1

StefanRijnhart: Thanks, merged!

astirpe (approved):
The travis failure is not caused by this PR.
Module tested, works fine. Code review ok.

astirpe:
Travis fails because of this:
```
l10n_nl_bsn/README.rst:16: [E7901(rst-syntax-error), ]  Definition list ends without a blank line; unexpected unindent.
```
I will fix it tomorrow.

StefanRijnhart:
Thanks @astirpe, that worked!

hbrunn:
@StefanRijnhart I wanted to solve the conflict using github's web editor, but it did something quite weird. Can you rebase correctly please?

StefanRijnhart:
You mean the merge commit? That's what Github does. I think we should learn to live with it

hbrunn:
I can live with the merge commit (even though I prefer not to have it), but now the PR is still stuck as unmergable

StefanRijnhart:
image
?

hbrunn:
ah, thanks, I set `Rebase and merge` as standard, and that's not possible here. Indeed didn't check the other options. Merge commit it is then.

Issue #162, opened 2018-03-23
XAF export not stored


CasVissers-360ERP:
When exporting a XAF audit for companies with a high number of journal entries the computation can take longer then for example the nginx time-out time or the employee's work time :) 

Can we attach generated XAF files to the object and add a list view so we can see and reload historic XAF files?

Thank you.
Cas

PR #171, opened 2018-06-28
[11.0][FIX] l10n_nl_xaf_auditfile_export: enable list of XAF files

astirpe:
When exporting a XAF audit for companies with a high number of journal entries the computation can take a long time. By default the generated XAF files are attached to the `xaf.auditfile.export` object and they remain like that. The problem is that those objects created in the past are not easily accessible anymore.

This is a simple proposal to enable the list view of the XAF exports, so that the user is able to access the history of the XAF files previously generated.

Fixes https://github.com/OCA/l10n-netherlands/issues/162 for V11.

cc @CasVissers

### DIFF
--- a/l10n_nl_xaf_auditfile_export/__manifest__.py
+++ b/l10n_nl_xaf_auditfile_export/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "XAF auditfile export",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.1",
     "author": "Therp BV, Odoo Community Association (OCA)",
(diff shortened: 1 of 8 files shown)


CasVissers-360ERP (approved):
Change tested in v10/v11 fixes the issue.

hbrunn:
I don't really understand commit 3c45b758d4a158089c60c3c1e3ebee775a866d3e, can you elaborate?

astirpe:
@hbrunn that commit was for me, just to see how the readme will look like. I just removed that commit. Sorry for the noise.

PR #174, opened 2018-06-30
[9.0][FIX] l10n_nl_xaf_auditfile_export: cast config param to int

astirpe:
Porting of #134 to V9


astirpe:
Includes fix of readme (l10n_nl_bsn) in order to make travis green

PR #179, opened 2018-07-16
[WIP][ADD] letsencrypt_transip_nl module

daramousk:
@hbrunn


letsencrypt_transip_nl/README.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Those fields are:
+    DNS Provider: Set this to TransIP
+    Login: Set this to the username you use to login to transip.nl
+    Key: Set this to the private key given to you by tranip.nl in order to use their APIs.
+    URL: Set the URL of the APIs. Usually api.transip.nl
+
+
+Usage
+=====
```
hbrunn: you can delete this section if there's nothing sensible to write

letsencrypt_transip_nl/README.rst
```diff
@@ -0,0 +1,86 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
```
hbrunn: for new modules, use [split readme files](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)

letsencrypt_transip_nl/models/letsencrypt.py
```diff
@@ -0,0 +1,45 @@
...
+    from transip.service.objects import DnsEntry
+    from transip.service import DomainService
+except ImportError as err:
+    _logger.debug(err)
+
+
+class Letsencrypt(models.AbstractModel):
+    _inherit = 'letsencrypt'
+
+    def letsencrypt_transip_support(self, challenge, domain):
```
hbrunn: you'll have to rename this function depending on what you do with https://github.com/OCA/server-tools/pull/1194/files#r230787313

letsencrypt_transip_nl/tests/test_letsencrypt_transip_nl.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
+
+
+class TestLetsencryptTransipNl(TransactionCase):
+    def test_letsencrypt_transip_nl(self):
```
hbrunn: If you plan the tests of the base module accordingly, you can here simply inherit those tests and rerun with a mock version of the transip library

letsencrypt_transip_nl/README.rst
daramousk: Not sure if I understand that readme file. Basically the README.rst that exists on the module's top directory is now removed and in its place we have a readme folder with various rst files as shown in the `maintainer-tools`?

letsencrypt_transip_nl/tests/test_letsencrypt_transip_nl.py
daramousk: Ok, but what is there to test in this module though besides raising coverage?

hbrunn: you mock the transip library and check it's called with the correct values

letsencrypt_transip_nl/README.rst
hbrunn: yes, in a `readme` subfolder. And then you can call https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py (this needs to be installed to generate the correct callables) to actually generate the readme to be sure it looks the way you want

l10n_nl_letsencrypt_transip_nl/models/letsencrypt.py
```diff
@@ -0,0 +1,51 @@
...
+        """ This function will be called by _respond_challenge_dns in order to
+            create the DNS record on transip.nl.
+        """
+        ir_config_parameter = self.env['ir.config_parameter']
+        login = ir_config_parameter.get_param('letsencrypt_transip_login')
+        key = ir_config_parameter.get_param('letsencrypt_transip_key')
+        url = ir_config_parameter.get_param('letsencrypt_transip_url')
+        token = base64.urlsafe_b64encode(challenge.token)
+        dns_entry = DnsEntry('_acme-challenge', 86400, 'TXT', token)
+        with NamedTemporaryFile() as f:
```
hbrunn: there seems to have been an update of the library so that all this isn't necessary any more: https://github.com/benkonrath/transip-api/blob/master/transip/client.py#L68

Currently this breaks, take this as a good argument why you should pass keyword parameters by name - had you done this, it would have kept working. This is good python style and makes your code resilient to slight API changes like this one.

l10n_nl_letsencrypt_transip_nl/tests/test_letsencrypt_transip_nl.py
```diff
@@ -0,0 +1,67 @@
...
+        )
+
+    @patch(
+        'odoo.addons.letsencrypt.models.letsencrypt.client'
+    )
+    @patch(
+        'odoo.addons.l10n_nl_letsencrypt_transip_nl.models.letsencrypt.'
+        'DomainService',
+    )
+    def test_invocation(self, domain_service, client):
```
hbrunn: seems it makes sense to write the tests of the letsencrypt module in a way that you can simply inherit from it (or maybe some common base class that both test classes inherit) so that you have to do the busy work only once

l10n_nl_letsencrypt_transip_nl/models/letsencrypt.py
```diff
@@ -0,0 +1,51 @@
...
+    def _respond_challenge_dns_transip(self, challenge, domain):
+        """ This function will be called by _respond_challenge_dns in order to
+            create the DNS record on transip.nl.
+        """
+        ir_config_parameter = self.env['ir.config_parameter']
+        login = ir_config_parameter.get_param('letsencrypt_transip_login')
+        key = ir_config_parameter.get_param('letsencrypt_transip_key')
+        url = ir_config_parameter.get_param('letsencrypt_transip_url')
+        token = base64.urlsafe_b64encode(challenge.token)
+        dns_entry = DnsEntry('_acme-challenge', 86400, 'TXT', token)
```
hbrunn: set the lowest [TTL](https://en.wikipedia.org/wiki/Time_to_live) the API accepts (the UI gives 60 seconds) - this decides how fast other nameservers invalidate old records. Especially for testing, a day is way too long

hbrunn (changes requested):

NL66278:
This PR has been superseded by #259

PR #182, opened 2018-07-24
[11.0][IMP] l10n_nl_xaf_auditfile_export: ease extensions

astirpe:
With this PR I'm proposing some small changes to ease the extention of `l10n_nl_xaf_auditfile_export` module.

- [x] extract method `_get_auditfile_template()`, to allow to extend this module by modules making use of different qweb templates (forward port of https://github.com/OCA/l10n-netherlands/pull/78);
- [x] make the computed field `auditfile_name` stored, to allow the old generated exports to keep their original name;
- [x] add group in view, to easily allow other modules to add extra options;
- [x] add basic tests to increase the coverage.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/__manifest__.py
+++ b/l10n_nl_xaf_auditfile_export/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "XAF auditfile export",
-    "version": "11.0.1.0.1",
+    "version": "11.0.1.1.0",
     "author": "Therp BV, Odoo Community Association (OCA)",
     "license": "AGPL-3",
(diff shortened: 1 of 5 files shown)


tarteo (changes requested):
I see this will make it easier if you want to replace the whole template.
Can you increase coverage, otherwise :+1:

astirpe:
Yes, this module does not contain tests yes. I could add some tests in this PR.

tarteo (approved):

RoelAdriaans (approved):

StefanRijnhart (approved):

astirpe:
Thanks!

PR #184, opened 2018-07-30
[10.0][REF] l10n_nl_postcodeapi: remove obsoleted fields_view_get()

astirpe:
In Odoo V10 the ORM `@api.onchange` methods completely replaced the `on_change="..."`  definitions on views.
For this reason the `fields_view_get()` method present in this module does not have any effect anymore and could be removed.


CasVissers-360ERP (approved):
Functional test ok.

StefanRijnhart (approved):

PR #186, opened 2018-08-01
[11.0][IMP] l10n_nl_xaf_auditfile_export: export auditfiles compatible with Unit4

astirpe:
This module extends module `l10n_nl_xaf_auditfile_export`: it
allows to export the XAF audit files in a format that is accepted by Unit4.

The Unit4 system expects a value for `<xsd:element name="docRef" ..>` of
maximum 35 characters, infringing the official standard. In case the value
exceeds 35 characters, an error is raised by Unit4 while importing the audit file.
By installing this module, the `docRef` value is simply truncated to 35 characters
so that the auditfile can be successfully imported into Unit4.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/__manifest__.py
+++ b/l10n_nl_xaf_auditfile_export/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "XAF auditfile export",
-    "version": "11.0.1.2.1",
+    "version": "11.0.1.3.0",
     "author": "Therp BV, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Accounting & Finance",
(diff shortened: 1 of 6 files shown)


StefanRijnhart:
Did you consider making this an optional setting in the core XAF module?

astirpe:
@StefanRijnhart that's an option, yes. If you all agree I can include the functionality in the core module.

StefanRijnhart:
I think I would prefer that. @hbrunn, @NL66278, any opinion on the matter?

NL66278:
I would also prefer to make this an option. Has Unit4 been made aware that their restriction violates the standard?

hbrunn:
+1 for option

astirpe:
Functionality is now included in the core XAF module, available as an optional setting

NL66278 (approved):
:+1:  LGTM

StefanRijnhart (approved):

yung-wang (approved):
Works fine

NL66278:
I think the Work In Progress label can be removed now, and the PR approved.

PR #189, opened 2018-08-13
[11.0][MIG] l10n_nl_country_states

astirpe:
This is the porting of `l10n_nl_country_states` to V11.


pedrobaeza:
Isn't interesting for you to switch to `base_location_geonames_import` module + maybe a little adaptation like the one in Spanish localization: https://github.com/OCA/l10n-spain/tree/11.0/l10n_es_toponyms ?

astirpe:
@pedrobaeza the `base_location_geonames_import` seems to me not fitting very well for the Dutch localization.
The number of provinces is very small in the Netherlands. On the other side, we need to track the postcode ranges per province, which seems to me an information that is not present in the geonames.

pedrobaeza:
OK, understood.

hbrunn (changes requested):
I think your changes will yield unexpected results for partial writes, that was my reasoning for doing it in a somewhat convoluted way. When we write a zip code (and only that) on some partner in a non-Dutch country, and the zip code happens to be in our list, we'll write the state, right?

Further, can you share the code you used to combine intervals or was that some manual post processing?

astirpe:
@hbrunn you're right! I will restore the previous behavior. Thanks!

astirpe:
I restored the old behavior, adding tests to cover the case of writing the zip (and only the zip).
I updated the script with the one used for combining the intervals. The script assumes that the records of the .xls are sorted by postcode number.

astirpe:
Commit https://github.com/OCA/l10n-netherlands/pull/189/commits/f35ac00c2a452e89e373516363f7fbdcd3dfa374 is meant to avoid conflicts with tests of `l10n_nl_postcodeapi`.
It's a ugly but quick fix. If you have better ideas just let me know :)

hbrunn (approved):
thanks. I think we can also just have the postcode module depend on this one, but the changed test is fine with me too.

tarteo (approved):

astirpe:
Can this one be merged?

PR #191, opened 2018-09-04
[FIX][8.0] demo data should not be regular data

thomaspaulb:


l10n_nl_xaf_auditfile_export/__openerp__.py
```diff
@@ -13,6 +13,9 @@
         'base',
         'account',
     ],
+    "demo": [
+        "demo/res_partner.xml"
+    ],
     "data": [
         "demo/res_partner.xml",
```
astirpe: I think you forgot to remove this line inside "data" ?

astirpe (changes requested):
Could you update the version number also?

l10n_nl_xaf_auditfile_export/__openerp__.py
thomaspaulb: Very true, fixed now.

astirpe (approved):
Runbot failure doesn't seem to be related with this PR

tarteo (approved):

PR #199, opened 2018-10-01
[12.0][MIG] l10n_nl_bsn

astirpe:


espo-tony (approved):
LGTM :+1:

l10n_nl_bsn/README.rst
```diff
@@ -0,0 +1,91 @@
...
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+======================================
+Burgerservicenummer (BSN) for Partners
+======================================
+
+This module adds the BSN (Burgerservicenummer) field on partner forms.
+
+The field is visible when the flag is_company is false.
```
espo-tony: This line is not true anymore, as flag is_company was removed in v12. The field is visible, instead, when "Individual" radio button is checked.

l10n_nl_bsn/README.rst
```diff
@@ -0,0 +1,91 @@
...
+
+* User must belong to the "HR Officer" group
+
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Open a form of a contact, eg.: a person (uncheck the flag "Is a Company?")
```
espo-tony: As above, this line is not accurate anymore

l10n_nl_bsn/models/res_partner.py
```diff
@@ -0,0 +1,68 @@
...
+            if partner.bsn_number:
+                # properly format the entered BSN
+                partner.bsn_number = bsn.format(partner.bsn_number)
+
+                # check is valid, otherwise display a warning
+                warning = partner._warn_bsn_invalid()
+
+                # search for another partner with the same BSN
+                args = [('bsn_number', '=', partner.bsn_number),
+                        ('name', '!=', partner.name)]
```
espo-tony, line 37: Why did you add a check on the name within the search domain?

l10n_nl_bsn/models/res_partner.py
```diff
@@ -0,0 +1,68 @@
...
+
+                # refine search in case of multicompany setting
+                if partner.company_id:
+                    args += [('company_id', '=', partner.company_id.id)]
+                other_partner = self.search(args, limit=1)
+
+                # if another partner exists, display a warning
+                if other_partner:
+                    warning = other_partner._warn_bsn_existing()
+        return {'warning': warning, }
```
espo-tony, line 47: If the BSN is not well formatted AND is already in use, you'd get a warning only for the latter, as the former value of the variable would be overwritten.
Is that the wished behavior? Wouldn't be better to have both issues signaled at once?

espo-tony (changes requested):
I tested the module and it works fine. Still I have some lesser changes to request and some questions.

l10n_nl_bsn/README.rst
astirpe: This is already fixed in file `readme/DESCRIPTION.rst` (https://github.com/OCA/l10n-netherlands/pull/199/files#diff-351c10adac39025dad8b1371e573db34R3). The `README.rst` will be automatically overwritten.

l10n_nl_bsn/README.rst
astirpe: Done in https://github.com/OCA/l10n-netherlands/pull/199/files#diff-de53f2d9eb0ad4e03a14910110c926dfR3

l10n_nl_bsn/models/res_partner.py
astirpe: Because you may have multiple records representing the same partner. For example, if 2 records of _PartnerA_ are present: you may want to accept both records of _PartnerA_  having the same `bsn_number`.

l10n_nl_bsn/models/res_partner.py
astirpe: I assumed that the warning "already in use" has the priority over the "not well formatted" one. You would likely check the "already in use" case first, in case both conditions happen.

l10n_nl_bsn/models/res_partner.py
astirpe: See also https://github.com/OCA/l10n-netherlands/pull/43

espo-tony: Ok. Thanks for your reply.

l10n_nl_bsn/models/res_partner.py
espo-tony: If it's the wished behavior that's fine by me. Thanks

espo-tony (approved):
:+1:

PatrickMallee (approved):
works fine

l10n_nl_bsn/README.rst
```diff
@@ -0,0 +1,91 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
```
tarteo: Can you generate the README.

tarteo (changes requested):

tarteo (approved):
Thanks!

astirpe:
With latest commit I'm proposing myself as maintainer of this module

PR #201, opened 2018-10-01
[12.0][MIG] l10n_nl_postcode

astirpe:


l10n_nl_postcode/models/res_partner.py
```diff
@@ -0,0 +1,51 @@
...
+            'title': _('Warning!'),
+            'message': msg % self.zip,
+        }
+        return warning
+
+    @api.multi
+    def _l10n_nl_postcode_check_country(self):
+        self.ensure_one()
+        country = self.country_id
+        if not country or country != self.env.ref('base.nl'):
```
espo-tony, line 32: The first part of the `if` condition is not useful, as it's implicitely included in the second part. Moreover, you could also write the whole check more compactly:

`if self.country_id != self.env.ref('base.nl'):`

This should largely suffice

espo-tony (commented):
Just a little note, otherwise it LGTM

l10n_nl_postcode/models/res_partner.py
astirpe: I would keep it as it is. It's meant to handle the case that 'base.nl' is not present in the system for some reasons.

espo-tony: Fine.

espo-tony (approved):

PatrickMallee (approved):
Works fine

l10n_nl_postcode/models/res_partner.py
tarteo, line 32: This can be ```return False if not country or country != self.env.ref('base.nl') else True``` or you can reverse the statement and just return it. But's not important

tarteo (approved):

l10n_nl_postcode/models/res_partner.py
astirpe: @tarteo I would like to keep it as it is. Thanks for the review!

PR #217, opened 2019-01-22
[12.0][MIG] l10n_nl_tax_statement_icp

astirpe:
Depends on:

- [x] https://github.com/OCA/l10n-netherlands/pull/203
- [x] #221


CasVissers-360ERP:
@astirpe 
Why do you hide the ICP statement when the tax statement is in draft?
`attrs="{'invisible':[('state','=','draft')]}">`

There is an update button available?

astirpe:
@CasVissers 
I don't remember anymore why did I make it this way, probably for simplicity or performance reasons.
Does it work fine if removing `attrs="{'invisible':[('state','=','draft')]}"`? Otherwise I need to find some time to handle that case correctly.

CasVissers-360ERP:
@astirpe It's because you only calculate the ICP when doing `@api.multi
    def post(self):`.  (self._compute_icp_lines())
The main reason why I don't like this concept is that VAT's for partners might be missing. A user then books, has to reset to draft and then recalculate the ICP statement. Seems illogical to me. 

There are multiple improvements possible, ie:
- recalculate button
- check if all partner's have a VAT before booking. (like this one most)

Anyhow. It's an improvement the module is functionally OK. Also did a small technical review.

CasVissers-360ERP (approved):

CasVissers-360ERP (approved):

StefanRijnhart (approved):

PR #216, opened 2019-01-07
[10.0][IMP] l10n_nl_account_tax_unece: add support for tax codes 9%

astirpe:
This is required after Odoo 10.0 commit https://github.com/odoo/odoo/commit/9ef7802520d9bdffd85861d31eaa28b5b094c39a

PR #222, opened 2019-02-18
[FIX] suppress extra space if we've just a firstname

hbrunn:
for partners without a lastname, we generate an extra space which can be hurtful in tests that do comparisons with full partner names


NL66278 (approved):
:+1: LGTM

NL66278:
Travis error is unrelated

astirpe:
Travis error is fixed by https://github.com/OCA/l10n-netherlands/pull/216

hbrunn:
@astirpe thanks, rebased into greenness

astirpe (approved):

PR #224, opened 2019-03-18
[ADD] Adds l10n_nl_privacy

daramousk:


rafaelbn:
Hello, we have the same requirement in Spain. Business must maintain 5 years the documentation.

Could we manage this in a more general approach like https://github.com/OCA/data-protection ?

Maybe in France or other countries is the same?

Just question, not blocking :smile

daramousk:
Hello there,

I have discussed internally with @hbrunn about putting the generic GDPR requirements on the `data-protection` repository and the country specifics on the `l10n-*` repositories.

Do you have a different approach?

hbrunn:
if it turns out this is not really nl specific, it makes sense to add this as config option in the base module. I'm still in favor of making the whole thing fully configurable with no code involved at all, but that makes the module much more complex than it currently is

hbrunn:
is this still relevant?

NL66278:
@hbrunn I think it can be closed. Not used by any of Therp's customers anyway.

PR #230, opened 2019-05-07
[12.0][MIG] l10n_nl_account_tax_unece

astirpe:


l10n_nl_location_nuts/tests/test_l10n_nl_location_nuts.py
```diff
@@ -10,11 +10,15 @@ class TestNlLocationNuts(TransactionCase):
     def setUp(self):
```
NL66278, line 10: The changes to this module should not be here I think.

l10n_nl_account_tax_unece/post_install.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, SUPERUSER_ID
+
+
+def set_unece_on_taxes(cr, _):
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        for company in env['res.company'].search([]):
+            if company.country_id and company.country_id != env.ref('base.nl'):
```
NL66278: Would it not be more safe to have this only work for companies that have country_id explicitly set to the Netherlands?

astirpe: Right, thanks!

l10n_nl_account_tax_unece/post_install.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, SUPERUSER_ID
+
+
+def set_unece_on_taxes(cr, _):
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        for company in env['res.company'].search([
+            ("country_id", "=", env.ref("base.nl").id),
```
NL66278: country_id on res.company is a non-stored field, leading to the travis errors. However searching on partner_id,country_id should solve this. Strangely enough when I use an odoo shell to directly search on country_id in res.company, like in your code, I do get the expected result as well... Even after checking that country_id is indeed not stored.

NL66278 (approved):
:+1: LGTM

Issue #231, opened 2019-05-15
l10n_nl_bsn: searching on BSN without dots


thomaspaulb:
I have seen that the `l10n_nl_bsn` module reformats the BSN by placing dots between the numbers in certain places, eg. `123.45.678` instead of `12345678`. However, when reading for example CSV files with BSN data and trying to match on it, the data will often be without those dots and thus the matching/searching on BSN will not work without extra modifications. Searching on a number without dots will also not work well in the regular Odoo "advanced search".

To me, the dots are formatting and the BSN should not be stored with them; rather there should be something like a BSN widget that shows the formatted BSN but the stored number is still vanilla. This is IMO the "right" way and will also aid in searching and comparing.

Any opinions on this?

astirpe:
Yes, I agree.
Alternatively we could implement a [search method](https://github.com/odoo/odoo/blob/12.0/odoo/fields.py#L157) for the bsn field. This will fix the Odoo "advanced search" and probably will also fix the importing by CSV.

thomaspaulb:
True, but such a search method should then be accompanied by either of these three:

1. A migration script that adds dots to existing BSN's, very strict enforcing that dotless BSN's don't enter the database and searching on BSN's with dots only
2. Some Postgres trick that allows a `where` search on a transform of the BSN value to something without dots (i dont know if this exists)
3. A stored, indexed computed field without dots that the search operates on

astirpe:
@thomaspaulb the above three points are not necessary IMO. The search method must allow the search on both dotted and not-dotted records, so no need of migration scripts or extra computed fields.

thomaspaulb:
Ah yes, my mind missed the option of searching for dotted as well as non-dotted. That should also do it.

PR #232, opened 2019-05-23
[12.0][FIX] l10n_nl_bsn: searching on BSN without dots

astirpe:
Module `l10n_nl_bsn` reformats the BSN by placing dots in the number.
This PR allows the searching on a number even without specifying the dots.

Fixes https://github.com/OCA/l10n-netherlands/issues/231 for V12.

### DIFF
--- a/l10n_nl_bsn/models/res_partner.py
+++ b/l10n_nl_bsn/models/res_partner.py
@@ -1,9 +1,10 @@
-# Copyright 2016-2018 Onestein (<https://www.onestein.eu>)
+# Copyright 2016-2019 Onestein (<https://www.onestein.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import logging
 
 from odoo import api, fields, models, _
+from odoo.osv import expression
 
 _logger = logging.getLogger(__name__)
 try:
@@ -17,33 +18,31 @@ class ResPartner(models.Model):
 
     bsn_number = fields.Char(
         string='BSN',
-        groups='hr.group_hr_user'
+        groups='hr.group_hr_user',
     )
 
-    @api.multi
     @api.onchange('bsn_number')
     def onchange_bsn_number(self):
         warning = {}
-        for partner in self:
-            if partner.bsn_number:
-                # properly format the entered BSN
-                partner.bsn_number = bsn.format(partner.bsn_number)
+        if self.bsn_number:
+            # properly format the entered BSN
+            self.bsn_number = bsn.format(self.bsn_number)
 
-                # check is valid, otherwise display a warning
-                warning = partner._warn_bsn_invalid()
+            # check is valid, otherwise display a warning
+            warning = self._warn_bsn_invalid()
 
-                # search for another partner with the same BSN
-                args = [('bsn_number', '=', partner.bsn_number),
-                        ('name', '!=', partner.name)]
+            # search for another partner with the same BSN
+            args = [('bsn_number', '=', self.bsn_number),
+                    ('name', '!=', self.name)]
 
-                # refine search in case of multicompany setting
-                if partner.company_id:
-                    args += [('company_id', '=', partner.company_id.id)]
-                other_partner = self.search(args, limit=1)
+            # refine search in case of multicompany setting
+            if self.company_id:
+                args += [('company_id', '=', self.company_id.id)]
+            other_partner = self.search(args, limit=1)
 
-                # if another partner exists, display a warning
-                if other_partner:
-                    warning = other_partner._warn_bsn_existing()
+            # if another partner exists, display a warning
+            if other_partner:
+                warning = other_partner._warn_bsn_existing()
         return {'warning': warning, }
 
     @api.multi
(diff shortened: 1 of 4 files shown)


astirpe:
@thomaspaulb I'm hereby experimenting the searching on BSN without dots. I'm sure the solution can be refined and improved but for the moment this is a quick fix. Would you give it a look?

l10n_nl_bsn/tests/test_l10n_nl_bsn.py
```diff
@@ -48,3 +48,43 @@ def test_03_bsn_another_partner(self):
...
+        res = self.env['res.partner'].search([
+            ('bsn_number', 'ilike', '00.00.009'),
+        ])
+        self.assertTrue(res)
+        self.assertEqual(len(res), 1)
+        res = self.env['res.partner'].search([
+            ('bsn_number', 'ilike', '00.00.00'),
+        ])
+        self.assertTrue(res)
+        self.assertEqual(len(res), 1)
```
thomaspaulb, line 83: The equal test already implies the True test?

l10n_nl_bsn/tests/test_l10n_nl_bsn.py
```diff
@@ -48,3 +48,43 @@ def test_03_bsn_another_partner(self):
...
+        res = self.env['res.partner'].search([
+            ('bsn_number', 'ilike', '00.00.009'),
+        ])
+        self.assertTrue(res)
+        self.assertEqual(len(res), 1)
+        res = self.env['res.partner'].search([
+            ('bsn_number', 'ilike', '00.00.00'),
+        ])
```
thomaspaulb, line 82: This works, but supporting both `ilike 00.00.00` and `ilike 000000` at the same time will be difficult unless we can be sure that the stored value has no dots...

l10n_nl_bsn/models/res_partner.py
```diff
@@ -66,3 +67,30 @@ def _warn_bsn_existing(self):
...
+                ])
+                if bsn.is_valid(bsn_number):
+                    bsn_format = bsn.format(bsn_number)
+                    bsn_domain = expression.OR([
+                        bsn_domain,
+                        [('bsn_number', operator, bsn_format)],
+                    ])
+                res_domain += bsn_domain
+            else:
+                res_domain.append(domain)
```
thomaspaulb, line 96: Something like the below is probably a very bad idea for so many reasons, but at least will support that `ilike` search:

```
operator = domain[1]
try:
    bsn_number = domain[2].replace('.', '')
except AttributeError:
    bsn_number = domain[2]
sql = expression.expression([('bsn_number', operator, bsn_number)], self)).to_sql()
sql[0].replace('"res_partner"."bsn_number"::text', "replace(bsn_number, '.', '')")
self.env.cr.execute('select id from res_partner where ' + sql[0] % (sql[1],))
ids = self.env.cr.fetchall(...)
return super().search([('id', 'in', ids), offset=offset, limit=limit, order=order, count=count)])
```

l10n_nl_bsn/tests/test_l10n_nl_bsn.py
astirpe: Yes, I could simplify by removing the assertTrue

l10n_nl_bsn/tests/test_l10n_nl_bsn.py
astirpe: Yes, true if we want to be 100% correct. But does it worth it?

l10n_nl_bsn/models/res_partner.py
astirpe: I got the idea :smirk: but my doubt is: does it worth it?

thomaspaulb: Not now, unless a customer asks for it and keeps insisting and we cannot talk sense :-)

Let's place the limitation in README in "Known issues" and go with your improvement

thomaspaulb (approved):

thomaspaulb (approved):

l10n_nl_bsn/models/res_partner.py
```diff
@@ -66,3 +65,30 @@ def _warn_bsn_existing(self):
...
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        res_domain = []
+        for domain in args:
+            if (
+                    len(domain) > 2 and domain[0] == 'bsn_number' and
+                    isinstance(domain[2], str) and domain[2]
+            ):
+                operator = domain[1]
```
hbrunn, line 81: I'd suggest to do this only if `operator` is not in https://github.com/OCA/OCB/blob/12.0/odoo/osv/expression.py#L151 - otherwise we get false positives I think.

I also think it should be possible to turn off this functionality via some context key.

l10n_nl_bsn/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+* The 'ilike' search of a substring of a BSN number could not work properly in case the
+  entered BSN number does not contain dots while the corresponding BSN number stored in the
+  database contains dots.
```
hbrunn, line 3: Isn't this rather a changelog entry? So should go to `newsfragments`?

hbrunn (changes requested):

l10n_nl_bsn/readme/ROADMAP.rst
astirpe: This was meant to be a warning, explaining that the search would be not working properly in some particular cases.
In case of better solutions will be proposed, this warning could be removed one day.

hbrunn: isn't this the problem you fix here?

astirpe: Yes, quick fix but not fully working: https://github.com/OCA/l10n-netherlands/pull/232#issuecomment-495118680

hbrunn: ah, I missed the *substring* part. Clear now.

l10n_nl_bsn/models/res_partner.py
astirpe: Thanks. Done

hbrunn (approved):

Issue #237, opened 2019-08-29
Tax Statement and ICP: allow a fiscal unit to do a tax statement for all the child companies. For the ICP the Totals of access to the childs.


yung-wang:
According the Dutch Tax Authority, as a fiscal unit, you can do 1 tax statement for all companies including the child companies.

In the current module, you can only see your the tax statement of the company you're logged in. In Enterprise, you can do a statement of all the companies, by selecting them when you're logged in into the parent company. It would be nice to have the same functionality in this module.

https://www.belastingdienst.nl/wps/wcm/connect/bldcontentnl/belastingdienst/zakelijk/btw/hoe_werkt_de_btw/voor_wie_geldt_de_btw/fiscale_eenheid/

For the ICP, the fiscal unit is allowed write down the totals of the ICP of all the companies.
https://www.belastingdienst.nl/wps/wcm/connect/bldcontentnl/belastingdienst/zakelijk/btw/zakendoen_met_het_buitenland/goederen_en_diensten_naar_andere_eu_landen/opgaaf_icp/opgaaf_icp_door_fiscale_eenheden

Requirement:
- The same Tax codes labels for all the companies.
- Company must be in The Netherlands

astirpe:
The ICP reports must be created separately for each company in any case, right?

astirpe:
@yung-wang could you check https://github.com/OCA/l10n-netherlands/pull/238?

yung-wang:
fixed in https://github.com/OCA/l10n-netherlands/pull/238

PR #221, opened 2019-02-08
[12.0][FIX/IMP] l10n_nl_tax_statement: add past invoices in report

astirpe:
Fixes #219
Forward port of https://github.com/OCA/l10n-netherlands/pull/220

PR #238, opened 2019-09-13
[12.0][IMP] Tax Statement: multicompany fiscal unit

astirpe:
Fixes https://github.com/OCA/l10n-netherlands/issues/237

Depends on:

- [x] https://github.com/OCA/l10n-netherlands/pull/221

### DIFF
--- a/l10n_nl_tax_statement/models/account_tax.py
+++ b/l10n_nl_tax_statement/models/account_tax.py
@@ -89,3 +89,42 @@ class AccountTax(models.Model):
                 ('tax_ids', 'in', tax_ids)
             )
         return res
+
+    def get_move_line_partial_domain(self, from_date, to_date, company_id):
+        if not self.env.context.get('fiscal_entities_ids'):
+            return super().get_move_line_partial_domain(
+                from_date, to_date, company_id)
+        return [
+            ('date', '<=', to_date),
+            ('date', '>=', from_date),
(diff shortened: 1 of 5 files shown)


StefanRijnhart:
Does this still allow people to file their NL VAT declarations for each company individually?

astirpe:
In case a group of companies are configured as belonging to a fiscal unit, I assumed that it's possible to declare only one single global tax statement for the whole group. It means that the generation of the single NL VAT declarations for each company of the group is disabled. To create an individual VAT declaration for a single company, this company should be configured as not being part of the fiscal unit. This is what the module does.

StefanRijnhart:
Ah yes, I missed that part where the user has to select the child companies explicitely. Great feature!

StefanRijnhart (approved):

StefanRijnhart:
Please review @yung-wang as this was your feature request.

yung-wang:
tested and it works fine.

1. Create 3 companies. Link comp 1 and comp 2 to headoffice.
2. Create invoices in comp 1 and comp 2.

result
I get 1 total tax statement for all the companies.
In the individual companies, i get a separate ICP.

astirpe:
@yung-wang "Files changed" > "Review changes" > "Approve" ?

yung-wang (approved):
see previous comment

PR #244, opened 2019-10-21
[FIX] l10n_nl_tax_statement,l10n_nl_xaf_auditfile_export: remove redundant po file

Yenthe666:
As the `nl.po` file is fully translated and `nl_NL.po` is the same language we should remove the file.


pedrobaeza:
@astirpe @hbrunn can you please check this?

Yenthe666:
Note: this issue is also present in the 12.0 branch so if you'd like I can make you a second PR to fix this too.

hbrunn (approved):
thanks, the v12 PR will be appreciated too

astirpe:
@Yenthe666 the `nl_NL.po` contains differences compared to the `nl.po`.
Those differences were explicitly made to fix the NL translation.
So as far as I understand, this PR is not okay :-1: 
But please ask the Dutch translation team to review this PR.

@yung-wang @CasVissers

pedrobaeza:
I'm curious, @astirpe are there 2 regional variants, nl and nl_NL? In Spain, there are a lot of obsolete files that are named as es_ES.po instead of es.po, but this is only an historic reason. There shouldn't be both.

astirpe:
@pedrobaeza Please ask the translation team. As far as I remember, that file was placed there for some reasons. This is a translation issue, not a debate of how many .po files there should be. I'm not in the Dutch translation team, so please ask them instead.
Thanks.

Yenthe666:
Why wouldn't the original `nl.po` file be just updated then though? It makes no sense to copy over all the content from an original language in a second file to fix a very minor amount of issues. The original terms could just be modified and it could be fixed on any database by updating the terms? Unless I miss something here.. :)

Yenthe666:
& @pedrobaeza to answer you: yes and no. We have the Netherlands who use `nl` and we have Belgium where they either speak branch (`fr_BE`) or Netherlands (`nl_BE`). In theory `nl` and `nl_BE` can differ but a few years ago we more or less decided to drop `nl_BE` for Belgium and just go with `nl.po`. The differences between the Dutch part in Belgium and the Netherlands are extremely small and not worth the hassle of maintaining and managing translating twice.

astirpe:
Yenthe, yes. There should be some coordination between the Belgian-flemish translation team and the Dutch translation team. I cannot help you for that. But if you ping them I'm sure yo will find a good solution.

pedrobaeza:
@Yenthe666 we have that case as well with regional languages like Galician or Catalonian, but they have a differentiated language code (or at least the second part). My point is that `nl_NL.po` should be the same than `nl.po` according to ISO codification (one is the short form, and the other is the full one).

Yenthe666:
So, who can make this final call now? ;)

Yenthe666:
Friendly two-week reminder.

Yenthe666:
Forget it, I'm not putting more resources into this. This is yet another clear reminder of why I don't collaborate more with OCA.

Yenthe666:
I see this one aged well. Still in V11 & not a single comment or TLC anymore. The 'great' OCA.

Issue #225, opened 2019-04-03
XAF export no starting balance


krijn1981:
We runned a XAF file (the V9 version) through a XAF analyser. But it look like there is not starting balance data. I am not a technical guy but what I can see in the code we only talking about start- and end-date of the fiscal year.

But as properly every knows Odoo has left since V8 the concept of fixed periods.
So since V8 there is not opening period (0). Therefore it will not appear anymore in the Audit-file.

Can somebody confirm that my findings are match with the code.
I see call of start- and end date in the V9/10/11 versions. but not a the "creation" of a period 0/opening/starting balance.

As my understandings the XAF module is only fetching the mutations of the exported year.

Binformed:
Hi OCA, can the previous comment be confirmed?

If yes, will it be fixed as more and more clients are reporting the same issue.

I would be nice to at least receive a confirmation about the code.

Thanks in advance.

astirpe:
Here is the code you are mentioning https://github.com/OCA/l10n-netherlands/blob/9.0/l10n_nl_xaf_auditfile_export/views/templates.xml#L104-L106 and that is the confirmation that the starting balance is not yet implemented.

krijn1981:
Hi, yes i see @astirpe is there a planning to create this?
Currently with XAF file the account can not fully his checkup by his XAF/controlling software.

astirpe:
@krijn1981 I'm not aware about any plan to add the starting balance. You may ask the authors of the module to know whether someone developed this. If you want you could also try to contact the company I'm working for and arrange an agreement for the developments.

PR #247, opened 2019-12-03
[ADD] [10.0] l10n_nl_xaf_auditfile_export: opening balance calculation

thomaspaulb:
This PR implements the `openingBalance` tag as asked for in https://github.com/OCA/l10n-netherlands/issues/225.

I have backported some commits from 11.0 in an effort to backport the tests. I have yet to make specific tests for the "openingBalance" part, but at least the current tests show that the report generates correctly even with the new changes.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -26,6 +26,8 @@ import os
 import psutil
 import shutil
 from tempfile import mkdtemp
+from io import BytesIO
+import zipfile
 import time
 from datetime import datetime, timedelta
 from dateutil.rrule import rrule, MONTHLY
@@ -54,10 +56,15 @@ class XafAuditfileExport(models.Model):
     _inherit = ['mail.thread']
     _order = 'date_start desc'
 
-    @api.depends('name')
+    @api.depends('name', 'auditfile')
     def _compute_auditfile_name(self):
         for item in self:
             item.auditfile_name = '%s.xaf' % item.name
+            if item.auditfile:
+                auditfile = base64.b64decode(item.auditfile)
+                zf = BytesIO(auditfile)
+                if zipfile.is_zipfile(zf):
+                    item.auditfile_name += '.zip'
 
     @api.multi
     def _compute_fiscalyear_name(self):
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -20,18 +20,32 @@
 ##############################################################################
 import base64
 import collections
-from StringIO import StringIO
 from lxml import etree
+import logging
+import os
+import psutil
+import shutil
+from tempfile import mkdtemp
+import time
 from datetime import datetime, timedelta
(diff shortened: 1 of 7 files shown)


astirpe:
@krijn1981 @Binformed you should be interested in reviewing this!

astirpe:
@thomaspaulb you should rebase your branch in order to fix the failure of runbot and travis

thomaspaulb:
@astirpe Thanks for the pointer, hope it goes green now.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -216,6 +252,45 @@ def get_taxes(self):
...
+    @api.multi
+    def get_ob_lines(self):
+        '''return opening balance entries'''
+        self.env.cr.execute(
+            'select a.id, a.code, sum(l.balance) '
+            'from account_move_line l, account_account a, '
+            '     account_account_type t '
+            'where a.user_type_id = t.id '
+            'and a.id = l.account_id and l.date < %s '
+            'and (l.company_id=%s or l.company_id is null) '
```
StefanRijnhart, line 284: It should technically, logically and functionally not be possible to have lines without a company ;-)

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -216,6 +252,45 @@ def get_taxes(self):
...
+            'and t.include_initial_balance = true '
+            'group by a.id, a.code',
+            (self.date_start, self.company_id.id, ))
+        for result in self.env.cr.fetchall():
+            yield dict(
+                account_id=result[0],
+                account_code=result[1],
+                balance=result[2])
```
StefanRijnhart: Please `round` this `result[2]` here to prevent errors like `Element '{http://www.auditfiles.nl/XAF/3.2}amnt': [facet 'fractionDigits'] The value '0.300000000001' has more fractional digits than are allowed ('2').`

StefanRijnhart:
Thanks, great work! I validated the results on one of my setups. Just two small comments.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
thomaspaulb: You're right, it's all over the code though, eg. [here](https://github.com/OCA/l10n-netherlands/blob/10.0/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py#L226)

I was lazy to correct them all, so maybe it can be an improvement for later that can then be back/forward ported to all of them.

StefanRijnhart: Yes, alright

StefanRijnhart (approved):
Thanks for the updates! Code review and test.

astirpe (approved):
LGTM

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
astirpe: Forward-ported to V13 (see https://github.com/OCA/l10n-netherlands/pull/253) with removed `l.company_id is null`, as suggested.

PR #255, opened 2020-02-04
[UPD] List of ISO 3166-1 country codes

StefanRijnhart:


l10n_nl_xaf_auditfile_export/data/XmlAuditfileFinancieel3.2.xsd
```diff
@@ -862,7 +869,6 @@
 			<xsd:enumeration value="TZ"/>
 			<xsd:enumeration value="UA"/>
 			<xsd:enumeration value="UG"/>
-			<xsd:enumeration value="UK"/>
```
astirpe, line 865: Shouldn't deleted values be kept instead? For backwards compatibility, to avoid failures while exporting old records?

StefanRijnhart: To be honest, I don't think so. I don't think there never was a country with an ISO UK code (although it's common to confuse it with GB of course). And in other cases, it's countries that are renamed. The old codes are not valid anymore. It seems entirely possible that the tax authority won't accept these illegal country codes, so it's important to keep up to date. The change in Odoo is limited to changing the country code of one or two countries.

astirpe: Yes, I see.
I just checked: `AN` code is still [present in Odoo 10.0](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_country_data.xml#L56) although it was removed in 11.0+ with https://github.com/odoo/odoo/commit/bfbaaa7d1319d109018b4e77ff930dadb8211cee#diff-c935af95c0540fa947cf91a105585dd4

astirpe (approved):

CasVissers-360ERP (approved):
👍 LGTM! Thnx for the fix.

CasVissers-360ERP:
BTW I think the CurrencycodeIso4217 should also be updated.

CasVissers-360ERP:
[XmlAuditfileFinancieel3.2.xsd.zip](https://github.com/OCA/l10n-netherlands/files/4557647/XmlAuditfileFinancieel3.2.xsd.zip)
I updated the currency codes in attachment.

PR #258, opened 2020-04-06
[11.0][IMP] l10n_nl_postcodeapi: mock tests

astirpe:
This PR converts tests of module `l10n_nl_postcodeapi` in order to make use of offline mock tests instead of real connections to PostcodeAPI service.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ env:
     - LINT_CHECK="1"
     - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="l10n_nl_postcodeapi"
     - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1" EXCLUDE="l10n_nl_postcodeapi"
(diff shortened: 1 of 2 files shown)


NL66278:
Tests are failing on lines 148 and on the same assertion in 196, street_name in partner is not set to Claudius Prinsenlaan.

NL66278:
@astirpe just wondering on the split tests: will we still not get travis failures when updating other modules because of the tests on the postcode module failing while running the lines with INCLUDE?

astirpe:
@NL66278 I split the tests because I don't have at the moment a better solution.
Locally I'm not able to reproduce the failure in https://travis-ci.org/github/OCA/l10n-netherlands/jobs/671978487 and I don't understand what is causing that conflict. I spent some time to investigate without any luck.
Maybe I will find some extra time in the future, but my original intention with this PR was just to fix the broken branch and avoid other PRs to be blocked.

NL66278 (approved):
:+1: LGTM. We have to await the results of the split later. And maybe have an additional PR later if needed for the remaining failing tests in the postcode API.

hbrunn (approved):

PR #261, opened 2020-05-06
[13.0] l10n_nl_xaf_auditfile_export only posted moves

CasVissers-360ERP:
The XAF file should only used posted moves. Cancelled and draft moves should be skipped.

@astirpe


astirpe (approved):

NL66278 (approved):
:+1: LGTM

PR #263, opened 2020-07-28
[ADD] l10n_nl_mis_reports

hbrunn:
this adds the v12 version of [`account_financial_report_horizontal`](https://github.com/OCA/account-financial-reporting/tree/10.0/account_financial_report_horizontal) based on [`mis_builder`](https://github.com/OCA/mis-builder)


bosd (changes requested):
The report view seems to have some styling issues. As it is generating a way to wide report.
Screenshot from 2021-01-23 16-15-34

Don't know if this is the proper solution,
But disabling the `width:100%;` in `.o_form_view` seems to fix it.
Screenshot from 2021-01-23 16-16-29

l10n_nl_mis_reports/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2020 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Dutch MIS Builder templates",
+    "summary": "Profit & Loss / Balance sheet for the Netherlands",
+    "version": "12.0.1.0.0",
+    "license": "AGPL-3",
```
bosd, line 8: Please add category so it shows up corretly in odoo appstore and backend applist.
```"category": "Localization",```
or
```"category": "Localization/Netherlands",```

hbrunn:
@bosd your css patch is way too broad, but to come up with a fix myself I need to reproduce the issue, which I fail to do. Which browser/platform/resolution is this?

bosd:
It was in Firefox, Resolution 3840*2160

bosd:
Jusst tested in Chrome it is ok.
In Firefox version 86.0 the problem exists.
The view looks a bit skewed in chrome browser as well.

The problem seems to come from the ``table`` class assignment which sets the width to 100% in the 2 tables within the ```<div class="mis_builder_horizontal"> ```
Resolved this by overriding the stylesheet with by adding ```style="width:50%;"```

Tested it in both Firefox and chrome. Looks ok now.


Do you have an clue why the header shows the text ``default``

hbrunn:
should be fixed now

bosd (approved):
Thanks for the fix!!
LGTM!

NL66278 (approved):
:+1: LGTM

bosd:
@StefanRijnhart Can you activate the bot? :robot:

StefanRijnhart:
@bosd I am not comfortable merging code that I have not reviewed myself. My considerations wrt this PR are: either I am missing something, or there is nothing NL specific about it. It looks to me like generic financial reports with a horizontal layout. While traditionally the balance reporting in NL has had the horizontal layout, it just looks odd in an international context, so more modern companies might favor the vertical one nowadays.
Having generic financial mis_builder reports is great, but I'm not sure they should be hosted here. Having a horizontal version of them is great as well, but preferably it should be an option and it could be implemented in the mis_builder core or a generic mis_builder_horizontal module. There is a preliminary discussion on the mis_builder project but it seems to have stalled.
So I'll abstain for now, from either reviewing or merging this.

bosd:
Thanks for the clarification. Not sure how much it is depending on the dutch CoA. So I don't have an informed opinion if this this PR is better suited for the "account financial reporting" repo. 
(But I am looking forward to using a V13-V14 port :wink:  )

ayushin:
This is just great! A long missed addition to the Dutch chart of accounts. Love it!

PR #270, opened 2020-08-14
[13.0][ADD] l10n_nl_ubl

astirpe:
With this localization module, Odoo will be able to generate UBL XML documents
by identifying the Dutch commercial partner with:

- Kamer van Koophandel (KvK).
- Organisatie-identificatienummer (OIN).

In case OCA module `partner_coc` is installed, this module will handle the proper
value of KvK. Otherwise if the field KvK is defined in another module,
you should extend method ``_l10n_nl_ubl_get_kvk`` returning your custom KvK field.

In case OCA module `l10n_nl_oin` is installed, this module will handle the proper
value of OIN. Otherwise if the field OIN is defined in another module,
you should extend method ``_l10n_nl_ubl_get_oin`` returning your custom OIN field.


CasVissers-360ERP:
Nice addition, a few comments:

1. 
image
I think this the right way of showing the KvK in the UBL file.

2. 
image
It would be good to add the OIN number https://www.logius.nl/diensten/oin, it's an identification number required in invoices for Dutch government agencies.

astirpe:
Thanks @casvissers-360erp.
Is there already a module for adding the OIN field? Otherwise I could create one on the fly..

CasVissers-360ERP:
@astirpe 
No I don't think there is a module for the OIN number yet.

astirpe:
Odoo V14 will include a generic [account_edi_ubl](https://github.com/odoo/odoo/tree/master/addons/account_edi_ubl) module in the core. Not sure if OCA will extend it or will keep the actual UBL modules independent. Anyway I'm thinking to rename this module and propose the following schema for the new upcoming UBL modules in Dutch localization:

#### Dutch localization for OCA UBL modules

- `l10n_nl_base_ubl` (this module) extending `base_ubl`
- `l10n_nl_account_invoice_import_ubl` (for importing UBL) extending `account_invoice_import_ubl`
- and so on...

#### Dutch localization for Odoo V14+ UBL modules

- `l10n_nl_edi` extending `account_edi_ubl`

The latter resembles the Belgian UBL module [l10n_be_edi](https://github.com/odoo/odoo/tree/master/addons/l10n_be_edi), also included in the Odoo core.

---
@CasVissers-360ERP what is your opinion about all of this?

CasVissers-360ERP:
@astirpe 
I agree with the naming.

From what I see account_edi_ubl will only be able to import UBL XML's (not export). I think/hope OCA will keep the UBL modules independent.

thomaspaulb:
@astirpe I'm going to work on this as well, unfortunately for another version of Odoo, but I think 90% will be the same, so hopefully we can add on eachother's work.

I need to do more changes than just adding the KVK and/or OIN numbers however, I need to add more fields to satisfy the business rules as laid out by the `NLCIUS` standard:

https://www.forumstandaardisatie.nl/open-standaarden/nlcius

NLCIUS is also the reason why KVK and OIN numbers are added to the invoice, so it makes sense to add it to this module. But shouldn't it then be renamed to `l10n_nl_nlcius_ubl`, in the same way that we have for example `l10n_fr_chorus_ubl`?

https://github.com/OCA/l10n-france/tree/12.0/l10n_fr_chorus_ubl

thomaspaulb:
@astirpe Sorry, I missed your naming proposal. I agree with it, but the Odoo 14 discussion is a complicated one and better held here: https://github.com/OCA/edi/issues/246

As to my own problem, I realise it's better to place my work in a new module `l10n_nl_account_invoice_ubl` since it only deals with invoice generation. `nlcius` vs `nl` is an academic detail I guess, `nl` is shorter so I'll go with that.

astirpe:
@thomaspaulb let me know if you need help by my side, eg.: porting modules to different versions, renaming modules, etc...
At the moment I'm not working on this module, so feel free to apply the changes you need :)

oca_dependencies.txt
```diff
@@ -1,2 +1,3 @@
 account-financial-reporting
 partner-contact
+edi
```
thomaspaulb, line 5: `base_ubl` depends in turn on `account_tax_unece` which is in `community_data_files`. Should that also be included here?

astirpe: Travis and Runbot are green, so it seems not needed. Otherwise we would see an error on both while installing the module.

thomaspaulb:
@astirpe It seems tests are still green, was the coverage red before?

thomaspaulb:
Update regarding this: NLCIUS is almost same with PEPPOL, so I went with a generic PEPPOL module:

https://github.com/OCA/edi/pull/323

If we depend on that one here and stick on the [NLCIUS header](https://github.com/OCA/l10n-netherlands/pull/290/files#diff-9fe902f0f7b6f7e9fdab8604360f3c60f813e89b19d35371ca7a9a1db3702a90R40), I think we're very close to generating valid NLCIUS.

l10n_nl_base_ubl/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Onestein (<https://www.onestein.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Base UBL - Netherlands",
+    "version": "13.0.1.0.0",
+    "category": "Dutch Localization",
```
bosd, line 7: Please Change category so it shows up corretly in odoo appstore and backend applist.
```"category": "Localization",```
or
```"category": "Localization/Netherlands",```

astirpe:
Resolved conflicts but now I get duplicated commits

NL66278:
@astirpe Maybe save the branch under another name, create anew from fresh 13.0 and then cherrypick + squash the right commits, then force push?

NL66278:
Sorry for closing / reopening, just hit the wrong button...

bosd:
There is a superseding pr. This once can be closed now

PR #271, opened 2020-08-17
[13.0][ADD] l10n_nl_oin

astirpe:
This module adds the Dutch `Organisatie-identificatienummer (OIN)` field on partner forms.


NL66278:
Module looks good. Just asking: why not follow the same methodology as with the coc number, basing it on partner_identification?

astirpe:
@NL66278 I received a lot of complaining regarding the adoption of `partner_identification` by `partner_coc`. At the end they just needed one field.
See discussion here: https://github.com/OCA/partner-contact/pull/645

CasVissers-360ERP:
I agree partner_identification brings a lot of extra options which makes no sense if your only purpose is to add a reference field to res.partner.

yung-wang:
Agree also, partner_identification is way to complex for just a reference field. partner_identification make more sense for passports, ID card etc.... not just for a ID number that can be used in other modules.

CasVissers-360ERP (approved):

NL66278 (approved):
:+1: LGTM

PR #272, opened 2020-08-21
[12.0][MIG] l10n_nl_postcodeapi

janverb:
A module for completing addresses using https://www.postcodeapi.nu/.

This contains [some changes besides the migration](https://github.com/oca/l10n-netherlands/compare/e673d45..496114a). We no longer make trial requests (b8c9c80d1c8f506d794e8c616f30c3208c6ccffa) because that eats into the new monthly limit, `ormcache` is no longer used (fc9242a6869c1b4adcc80efcb841e368c461f8ae), and a few other things. Some of it might be worth backporting.

@NL66278


l10n_nl_postcodeapi/i18n/nl_NL.po
```diff
@@ -0,0 +1,50 @@
+# Translation of Odoo Server.
```
NL66278: This file should not be there. Only the nl.po file.

l10n_nl_postcodeapi/models/res_partner.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2013-2015 Therp BV <https://therp.nl>
```
NL66278: You can extend copyright to 2020

NL66278 (changes requested):
:+1: LGTM But need to delete the nl_NL.po file

l10n_nl_postcodeapi/tests/test_l10n_nl_postcodeapi.py
```diff
@@ -0,0 +1,202 @@
+# Copyright 2018-2020 Onestein (<https://www.onestein.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from odoo.modules.module import get_module_resource
```
astirpe: According to https://github.com/odoo/odoo/blob/12.0/odoo/modules/module.py#L225-L226, I think it's better to use `get_resource_path` instead of `get_module_resource`.

l10n_nl_postcodeapi/tests/test_l10n_nl_postcodeapi.py
```diff
@@ -0,0 +1,202 @@
...
+from odoo.tests.common import TransactionCase
+
+
+class TestNlPostcodeapi(TransactionCase):
+
+    def setUp(self):
+        super(TestNlPostcodeapi, self).setUp()
```
astirpe:
```suggestion
        super().setUp()
```

l10n_nl_postcodeapi/models/res_partner.py
```diff
@@ -0,0 +1,52 @@
...
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.model
+    def get_provider_obj(self):
+        apikey = self.env['ir.config_parameter'].sudo().get_param(
+            'l10n_nl_postcodeapi.apikey', '').strip()
+        if not apikey or apikey == 'Your API key':
+            return False
+        from pyPostcode import Api
```
astirpe: Can you move this importing statement `from pyPostcode import Api` to the top of the file?

astirpe (changes requested):
Only code review, not tested.

astirpe (approved):

janverb:
Fixed, thank you.

I'm confused about codecov—it seems like [coverage for the module is at 100%](https://codecov.io/gh/OCA/l10n-netherlands/tree/76c2c013f5ea1c63a9627359c02bc652a9f1deeb) but it's still rejecting it.

NL66278 (approved):
:+1: LGTM and working in practice

NL66278:
I try once more with ocabot. Else will just merge it using the button.

PR #273, opened 2020-08-27
[UPD] new VAT tariff

thomaspaulb:
This could have been done in a much more future-proof way, eg. to get the percentage from the actual `account.tax.code` record being referenced, but as this is a legacy 8.0 report I'm not too bothered.


thomaspaulb:
@Koesj

gjotten (commented):
@thomaspaulb +1

gjotten:
Whoops this was already merged :)

PR #287, opened 2020-10-30
[12.0] [MIG] l10n nl partner salutation

NL66278:
Standard migration

- Also solved problem with form not showing labels for all fields;
- Adapted po/pot to new format (double underline between field and model name);
- Added active flag, so the salutations added by default (or others) can be deactivated.


astirpe (approved):
LGTM

l10n_nl_partner_salutation/models/res_partner.py
```diff
@@ -47,67 +47,63 @@ def get_salutation_name_format(self):
...
             infix=self.infix
         )
 
-    @api.depends(
-        'gender',
-        'title',
-        'use_manual_salutations',
-        'salutation_manual',
-        'salutation_address_manual'
-    )
```
thomaspaulb: Even though this is not a stored computed field, I was under the impression that it still needs a "depends" clause so that it can live-update in a form view in the way that "onchange" does. Is that true or am I mistaken?

l10n_nl_partner_salutation/models/res_partner.py
```diff
@@ -47,67 +47,63 @@ def get_salutation_name_format(self):
...
-            field itself has no value.
-            """
-            val = ''
-            if partner.gender in ('male', 'female'):
-                val = partner.title[field + '_' + partner.gender]
-            if not val:
-                val = partner.title[field] or \
-                    DEFAULT_PREFIX[field][partner.gender or 'unknown']
-            return (val + ' ') if val else ''
+    def _compute_salutation_address(self):
```
thomaspaulb, line 79: If I am correct, then this one also needs the `api.depends` clause.

l10n_nl_partner_salutation/models/res_partner.py
StefanRijnhart: Agreed. What do you think @NL66278?

l10n_nl_partner_salutation/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2017-2020 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Full salutation for partners, Dutch style',
+    'version': '12.0.1.0.0',
+    'author': 'Therp BV, Odoo Community Association (OCA)',
+    'category': 'Contact management',
```
bosd, line 7: Please Change category so it shows up corretly in odoo appstore and backend applist.
```"category": "Localization",```
or
```"category": "Localization\Netherlands",```
or
```"category": "Localization\Contact management",```

l10n_nl_partner_salutation/models/res_partner.py
NL66278: Agreed also. I added the depends back in.

PR #233, opened 2019-05-24
[11.0][FIX] l10n_nl_bsn: searching on BSN without dots

astirpe:
Backport of https://github.com/OCA/l10n-netherlands/pull/232 to V11

PR #293, opened 2020-11-24
[11.0][FIX/IMP] l10n_nl_kvk: secure connection to api.kvk with TLS certificate

astirpe:
This PR fixes the travis error on 11.0 branch which is broken at the moment.

Error is caused by module `l10n_nl_kvk`: connection with KVK API suddenly needs a TLS certificate to be compliant with recent changes made on KVK API.

See also comment https://github.com/OCA/l10n-netherlands/pull/233#issuecomment-732889398


astirpe:
Already included within https://github.com/OCA/l10n-netherlands/pull/233

PR #294, opened 2020-12-08
[8.0][FIX] Travis and Runbot

thomaspaulb:


l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
```diff
@@ -31,7 +31,7 @@ def test_l10n_nl_xaf_auditfile_export_default(self):
         zf = StringIO(b64decode(self.export.auditfile))
         with ZipFile(zf, 'r') as archive:
             contents = archive.read(archive.namelist()[0])
-        self.assertTrue(contents.startswith('<?xml '))
```
NL66278, line 34: Why these changes?

thomaspaulb: I see that the `l10n_nl_xaf_auditfile_export` test is failing on this line, but I cannot see what it generates instead of `<?xml `. And on my local the test is passing. An assertEquals would show me directly in Travis what is generated instead, so that I can fix it.

NL66278: OK, thanks for the explanation!

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -43,7 +43,7 @@ def _compute_auditfile_name(self):
             item.auditfile_name = '%s.xaf' % item.name
             if item.auditfile:
                 auditfile = base64.b64decode(item.auditfile)
-                zf = StringIO(auditfile)
+                zf = BytesIO(auditfile)
```
thomaspaulb, line 46: This is now in line with [10.0](https://github.com/OCA/l10n-netherlands/blob/10.0/l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py#L42)

l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
```diff
@@ -28,16 +28,18 @@ def setUp(self):
...
         self.export.write({'data_export': 'all'})
         self.export.button_generate()
         self.assertTrue(self.export.auditfile)
-        zf = StringIO(b64decode(self.export.auditfile))
+        zf = BytesIO(b64decode(self.export.auditfile))
         with ZipFile(zf, 'r') as archive:
-            contents = archive.read(archive.namelist()[0])
-        self.assertTrue(contents.startswith('<?xml '))
+            filelist = archive.filelist
+            contents = archive.read(filelist[-1]).decode('utf8')
```
thomaspaulb, line 44: Also now in line with [10.0](https://github.com/OCA/l10n-netherlands/blob/10.0/l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py#L42).

NL66278 (approved):
:+1: LGTM

thomaspaulb:
@hbrunn I see you are a PSC here.. could I look at you nicely?

PR #295, opened 2020-12-11
[13.0][UPD] Update res_bank_data.xml 

bosd:
Update Dutch BIC List as per 23-07-2020

### DIFF
--- a/l10n_nl_bank/data/res_bank_data.xml
+++ b/l10n_nl_bank/data/res_bank_data.xml
@@ -4,11 +4,11 @@
         <field name="name">ABN AMRO BANK N.V.</field>
         <field name="bic">ABNANL2A</field>
     </record>
-        <record id="ABNC" model="res.bank">
+    <record id="ABNC" model="res.bank">
         <field name="name">ABN AMRO CLEARING BANK N.V.</field>
         <field name="bic">ABNCNL2A</field>
     </record>
-        <record id="ADYEN" model="res.bank">
+    <record id="ADYEN" model="res.bank">
         <field name="name">ADYEN</field>
         <field name="bic">ADYBNL2A</field>
     </record>
@@ -32,7 +32,7 @@
         <field name="name">ASN BANK</field>
         <field name="bic">ASNBNL21</field>
     </record>
-        <record id="ATBA" model="res.bank">
+    <record id="ATBA" model="res.bank">
         <field name="name">AMSTERDAM TRADE BANK</field>
         <field name="bic">ATBANL2A</field>
     </record>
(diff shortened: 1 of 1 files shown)


CasVissers-360ERP (approved):
Functionally OK, code should be checked.

astirpe (approved):
Code review ok as it passes tests.

astirpe (changes requested):
Maybe is better to avoid the removal of old banks, otherwise it could create troubles in case they are used somewhere. Just adding new banks would be fine.

bosd:
In a new install, one would like to avoid importing old banks.
Can you explain where the deletion of the old banks lead to problems?

NL66278:
Updating a module should never lead to the deletion of user data.

CasVissers-360ERP:
Might be good to create a migration script to remove external ID's for old banks before initiating the upgrade.

astirpe:
I agree with @CasVissers-360ERP in case you want to remove old banks from the xml. That way it will avoid the removal of records in db while updating the module.
But do we really need to remove old banks?

CasVissers-360ERP:
I think removing old banks is a good idea for fresh installations, not a must. Lot's of new Odoo customers will still have old banks in their data. From consultant perspective I would love to tell customers when BIC number's are deprecated though. Would mean that Odoo is a an up-to-date system for the Dutch localisation.

astirpe:
For deprecating old banks on new installations, adding an `<field name="active" eval="False" />` would do the job

bosd:
I agree with @NL66278 `Updating a module should never lead to the deletion of user data.`

I did not test an upgrade situation. Would like to know how updates are handled.
( My reasoning behind this PR was a quick fix to use up to date data for new installations )

This data consists of only dutch banks.
In production databases there are also (user input) foreign banks from other countries.
The information of these foreign banks have never been in to this data set.

If the removal of old dutch bank data in this dataset would remove banks from a database in an update situation. 
Would that mean that the (user input data) from foreign banks would also be deleted?

@astirpe What do you propose to do now?


Instead of solving the data issue locally for each country, it would be really nice to have this handled globally by an ISO 9632 community-data-files module.
https://github.com/OCA/community-data-files/pull/13#issuecomment-194795003

astirpe:
> In production databases there are also (user input) foreign banks from other countries.

Those user input data are not a problem, they will not be involved in this process.


> If the removal of old dutch bank data in this dataset would remove banks from a database in an update situation.

Yes. I suspect that's what actually happens: even if noupdate=1,  the old banks are deleted from db while updating.

> @astirpe What do you propose to do now?

I would propose to simply add `<field name="active" eval="False" />` to old banks instead of removing them.

bosd:
Closed, this one, new PR in #299

PR #296, opened 2020-12-11
[12.0][IMP] Refactor to base postcodeapi on postcode check.

NL66278:
- Only retrieve postcode if this has a valid format;
- Refactor to make methods re-usable and re-use them;
- Provide a warning when postcode is not valid instead of ignoring problem.


bosd (changes requested):
Functional review:
When only a house number is provided and the postcode area is empty.
Error message “The Postcode you entered () is not valid.” is displayed. 

Whilst inputting same data for the sanbox api. This error message occurs:
Cannot validate postalcode (6545 CA), or postal code invalid for housenumber.
(the entered housenumber was 29)
This data should be available in the sandbox API.

Please update the readme file.
As this API is no longer free. 
Can you add which version of the API is used?

PR #297, opened 2020-12-12
[13.0][FIX] Delete nl_NL.po to remove duplicate translation

bosd:
Delete nl_NL.po to remove duplicate translation, so it will improve the translation coverage of this branch. (The main translation is in the other .po file)


astirpe:
A similar PR was already opened in the past and [this](https://github.com/OCA/l10n-netherlands/pull/244#issuecomment-544547150) was my comment.

NL66278:
@astirpe As far as I can see the duplicate file does not contain any translations where the Dutch would differ from nl.po. Actually it hardly contains any translations at all.

NL66278 (approved):
:+1: LGTM

astirpe:
I can see some differences of translations between `nl_NL.po` and `nl.po`. What I don't understand is why the translation team is not involved in this PR and why the differing translated terms in `nl_NL.po` cannot be moved to `nl.po`.
Anyway if you are happy with this PR just go on with the merge...

PR #298, opened 2021-01-12
[14.0][MIG] l10n_nl_tax_statement

astirpe:
Standard porting + `l10n_nl_tax_invoice_basis` merged within `l10n_nl_tax_statement`


CasVissers-360ERP:
@astirpe 
I was thinking why do we need l10n_nl_tax_invoice_basis. 

In **_get_unreported_move_domain** we use **l10n_nl_date_invoice** if **is_invoice_basis**.  But we already have the field **invoice_date** on account.move which we can use? I think it would be good to incorporate tax_invoice_basis in the base module.

astirpe:
@CasVissers-360ERP Yes, you are right! I will do it in the next days, unless you already implemented any solution :)

CasVissers-360ERP:
Great. I will test it :)

l10n_nl_tax_statement/views/l10n_nl_vat_statement_view.xml
```diff
@@ -0,0 +1,302 @@
...
+                                    options="{'related_start_date': 'from_date'}"
+                                    attrs="{'readonly': [('state','in',['posted','final'])]}"
+                                />
+                            </div>
+                        </group>
+                        <group name="extra_parameters">
+                            <field name="date_posted" />
+                            <field name="date_update" />
+                            <field name="currency_id" />
+                            <field name="target_move" />
```
CasVissers-360ERP: Might be good to show <field name="is_invoice_basis"/> here?

astirpe: This is a leftover, I'm removing it :+1:

StefanRijnhart:
If I understand correctly, Odoo supports cash basis, which means registering the tax when receiving payment (https://www.odoo.com/documentation/user/14.0/accounting/fiscality/taxes/cash_basis_taxes.html). It supports 'standard taxes' using the accounting date of invoices.

This module offers the option of using 'Invoice basis', which means: using the `date_invoice` of the invoice instead of the accounting date of the invoice. Whether it is necessary to file Dutch taxes in a legally correct manner depends on the processes and the discipline of the users. It is very well possible to apply 'Invoice basis' without this option.

The feature that is provided here is of course very useful for some companies, as it allows to post invoices after closing the period of the invoice while still maintaining consistency of the 'Invoice basis' tax declaration. So while I can see its usefulness, I am wondering if it is specific for the Dutch localization. Did you consider to discuss this in a greater context and maybe make this feature available in oca/account-financial-reporting? I think that should be more appropriate than merging it into this module.

astirpe:
@StefanRijnhart When I was discussing this matter internally with colleagues, I understood that 'Invoice basis' was a Dutch practice only, but of course we can open a RFC in `account-financial-reporting` and see if it gets interest.

What `l10n_nl_tax_invoice_basis` does is to modify the query used by `account_tax_balance` to use the `invoice_date` instead of `date`. Maybe it would be easier to add the option for 'Invoice basis' directly in `account_tax_balance`, what do you think?

StefanRijnhart:
Thanks for clarifying in our chat that this module is no longer based on `account_tax_balance`. As such, it is no problem to merge the functionality in `l10n_nl_tax_invoice_basis`. As discussed, theoretically there be differences between the views generated by account_tax_balance and the actual Dutch VAT declaration, but I'm slightly puzzled what the business case is to post invoices with an accounting date that is different from the invoice date so I'm not too fuzzed.*

Maybe you can add to the description of this module and the help attribute of the field 10n_nl_tax_invoice_basis in the configuration wizard what the option does exactly (i.e. report taxes on the invoice date rather than the accounting date). You need to update the documentation *anyway* because it still mentions this option as a separate module.

\* edit: talked to an accountant who informed me this is a common requirement wrt. servics for instance that are invoiced in the period after those services were executed. In this flow, costs are accounted for in that period and the vat in the period of the invoice date.

astirpe:
Thanks @StefanRijnhart, I fixed the readme

l10n_nl_tax_statement/views/l10n_nl_vat_statement_view.xml
```diff
@@ -0,0 +1,301 @@
...
+                                    widget="daterange"
+                                    nolabel="1"
+                                    class="oe_inline"
+                                    options="{'related_start_date': 'from_date'}"
+                                    attrs="{'readonly': [('state','in',['posted','final'])]}"
+                                />
+                            </div>
+                        </group>
+                        <group name="extra_parameters">
+                            <field name="date_posted" />
```
CasVissers-360ERP, line 116: Might be good to show l10n_nl_tax_invoice_basis somewhere in the form view?

CasVissers-360ERP (changes requested):
First feedback after invoice basis change.

l10n_nl_tax_statement/views/l10n_nl_vat_statement_view.xml
astirpe: @CasVissers-360ERP done!

l10n_nl_tax_statement/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,15 @@
+This module makes use of the tax tags (eg.: 1a, 1b, 1c, 1d, 2a...) as prescribed by the Dutch tax laws.
+
+If the default Odoo Dutch chart of accounts is installed (module ``l10n_nl``) then these tags are automatically present in the database.
+
+If a non-standard chart of accounts is installed, you have to manually create the tax tags and properly set them into the tax definition.
+The name of the tags must be formatted this way: "+1a (omzet)", "+1a (btw)", "-1a (omzet)", "-1a (btw)", "+2a (omzet)", "+2a (btw)", etc...
+
+When installing this module, by default the *Factuurstelsel* system is configured
+for your Company. To adopt the *Kasstelsel* instead, you need to disable the
+*Factuurstelsel* option.
```
StefanRijnhart: I'd suggest something like the following

```
This module provides an accounting setting *Invoice basis* (Factuurstelsel)
that is enabled by default. This option makes the *invoice date* leading for
the tax declaration of a period, rather than the accounting date, and
allows you to change the accounting date of incoming invoices to the period
to which the costs apply and still create a legally valid Dutch tax
declaration. Without this option, the entries on the tax statement are
collected using only their accounting dates (what Odoo calls *standard
taxes*).

This setting is not compatible with the Odoo *Cash basis* (Kasstelsel) setting
which leads to the creation of tax journal lines at the moment that a payment
is received, so when using *Cash basis* you need to disable the *Invoice
basis* setting to create a valid Dutch tax declaration using this module.
```

astirpe: Done, thanks!

StefanRijnhart: Thanks! I will test this module locally some time this week.

l10n_nl_tax_statement/views/l10n_nl_vat_statement_view.xml
CasVissers-360ERP: Another GUI thing, string="View base lines" is visible I think we only want the icon?

l10n_nl_tax_statement/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    l10n_nl_tax_invoice_basis = fields.Boolean(
+        string="NL Tax Invoice Basis",
+        related="company_id.l10n_nl_tax_invoice_basis",
+        readonly=False,
```
CasVissers-360ERP: Obsolete?

l10n_nl_tax_statement/models/l10n_nl_vat_statement_line.py
```diff
@@ -0,0 +1,127 @@
...
+    name = fields.Char()
+    code = fields.Char()
+
+    statement_id = fields.Many2one("l10n.nl.vat.statement")
+    currency_id = fields.Many2one(
+        "res.currency",
+        related="statement_id.company_id.currency_id",
+        help="Utility field to express amount currency",
+    )
+    omzet = fields.Monetary()
```
CasVissers-360ERP: Maybe better to switch to English? Omzet = Turnover

l10n_nl_tax_statement/models/l10n_nl_vat_statement_line.py
```diff
@@ -0,0 +1,127 @@
...
+    btw = fields.Monetary()
```
CasVissers-360ERP: Maybe better to switch to English? btw = vat

l10n_nl_tax_statement/views/l10n_nl_vat_statement_view.xml
```diff
@@ -0,0 +1,302 @@
...
+                                    <field name="format_omzet" invisible="1" />
+                                    <field name="format_btw" invisible="1" />
+                                    <field
+                                        name="omzet"
+                                        attrs="{'invisible': [('format_omzet', '=', False)],'readonly': [('is_readonly', '=', True)]}"
+                                    />
+                                    <button
+                                        name="view_base_lines"
+                                        type="object"
+                                        string="View base lines"
```
CasVissers-360ERP: String is visible in form view, we only want a button?

CasVissers-360ERP (approved):
Minor feedback. Functionality it is working correctly.

astirpe:
@CasVissers-360ERP done, thanks!

StefanRijnhart (approved):
Great work, as always! Can you squash the migration commits?

astirpe:
Thanks @StefanRijnhart ! Done.

StefanRijnhart:
Great, thanks!

/ocabot merge nobump

PR #299, opened 2021-01-15
[13.0][UPD] Update res_bank_data.xml

bosd:
Was unable to edit PR #295 
So opened this new PR.


bosd:
@CasVissers-360ERP and @astirpe 
Changes made as mentioned in the other PR.
Functionally tested. LGTM

Please review :pray:

StefanRijnhart:
Because it was suggested to set old banks to inactive, new installations still load the data (but inactive). I don't think we should burden new installations with an evergrowing list of obsolete data. An alternative solution is to delete the banks from the xml. Because the xml is/was recorded with noupdate="1", Odoo will not attempt to remove the banks from existing installations.

You now set noupdate to "0", I assume to ensure that Odoo will update existing installations with the changes from this PR but that is not what happens. Because the banks were loaded as noupdate data in the earlier version, they are marked as such in the database and these changes will not be applied on module upgrade.

I would keep the noupdate flag; remove obsolete banks from the data file; add a migration script to force-load the changes (for instance by using openupgrade's load_data method https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.load_data).

astirpe:
I just tried locally: when deleting old banks on xml with noupdate="1", Odoo does not remove old banks from db. So my previous comment was wrong, my apologies @bosd. For me removing old banks from xml would be fine, although not strictly necessary.
About a migration script for force-load the changes, adopting openupgrade would be causing an error while upgrading the modules in servers where openupgrade lib is not installed, we already seen in the past.

bosd:
During tests, there was an error with the setting noupdate set to 1.
In the first install it would populate the data. After upgrading it would load duplicates of the same data.
On future updates, it would not update the data as is expected from the "noupdate" flag.

From the other PR, @astirpe said
> I suspect that's what actually happens: even if noupdate=1, the old banks are deleted from db while updating.

Just tested this. When a record is deleted from the dataset, it wont be deleted from the db.


**Is the usage of the no update "1" flag the desired way?**
As this is a curated list, one would actually want to have the data updated. 
(Typo's, mergers and acquisitions on the banks end).

With the noupdate flag set to "0", as default on this module would mean that changes in the dataset are populated to the db on an upgrade.
Setting the inactive flag, on depreciated records would work for compatibility.
True, that new installs would load an ever growing list of inactive data.

**Would that be an problem?** As in this case it is not a lot of data. As in this case it is only dutch bank data.
It will likely be an situation which would change in the future. 
If the time happy moment comes when there is a module which pulls the data with an API from an external datasource.
With more advanced data handling features which would inform the odoo user when a BIC number becomes depriciated. (as mentioned by Cas)
BTW, the swiss are facing a similar situation with their bank information in https://github.com/OCA/l10n-switzerland/issues/514


The upside for updating the db data on an upgrade is that it keeps the workflow simple. No need for update scripts.

StefanRijnhart:
I see the changes to existing entries are only cosmetic, so you don't really need to keep them updated in existing setups IMHO.

 As you have confirmed, noupdate entries are not deleted when removed from the data. I think a list like this is the canonical use case for noupdate data.

StefanRijnhart:
If you insist on moving to remove the noupdate flag, you have to add a migration script to this PR resetting it for the existing entries in ir_model_data. I would deal with obsolete items in the following way: remove them from the noupdate data entirely and in a pre-migration script, delete the entry from ir_model_data so that the entry is not removed from existing setups.

StefanRijnhart:
@astirpe openupgradelib comes in very handy for advanced migration efforts like this. You could consider adding it as a module dependency in later versions of this module so that you can use its functionality later on in its lifecycle.

bosd:
> I see the changes to existing entries are only cosmetic

Lucky these are, some acquisitions. The data could even be enriched more with the adresses and such.
IMHO that would not really add more value for the user, but more data to maintain.


How do we reach consensus about moving this one forward?

To recap option:
1) Keep no update flag and remove the old records


2) Remove the no update flag, keep the old records in the file as inactive. Provide a migration script.

Dunno, if my skills are on par to provide a migration script.
As one of my first commits I was aiming at something simple :wink: 
Would'nt have thought it would become an advanced migration.

After the final changes, this file can be safely changed (cherry picked) in branches 10 until 13, right??

StefanRijnhart:
A migration script involves:
- A manual module version number increment (minor, second digit from the right)
- A subdirectory in the module `migrations/<VERSION>`
- In this case, a file in that subdirectory `pre-migration.py`
- In that file, a method `def migrate(cr, version):`
- In that method, a query

```
UPDATE ir_model_data SET noupdate = FALSE
WHERE module = 'l10n_nl_bank' AND model = 'res.bank'
```

Here is an example of a migration script https://github.com/OCA/l10n-netherlands/blob/11.0/l10n_nl_tax_statement/migrations/11.0.2.0.0/post-migrate.py but that is a post-script in which you can fetch an env. In a pre-script you need to make to with the cursor.

StefanRijnhart:
Changes can be cherrypicked but with altered version numbers.
BTW I still vote for the noupdate="1" option, but that's just me ;-)

bosd:
@StefanRijnhart  Thanks for the very good write-up.
Decided to go with your proposed solution.

> I would keep the noupdate flag; remove obsolete banks from the data file; add a migration script to force-load the changes

Managed to create a script which does not crash the server :smile:
But functionally it is doing the opposite from the objective.
It does not update existing records. It removes entries from the db which are removed from the datafile. :(

(openupgrade doc's don't describe the behavior of the update routine. My assumption was that it does only update. Func Test turned out it also deletes.)

please help :pray: 

pre-migrate.py

```
"""
The banks have been created with the no update tag.
Change it to update true to update the bank list.

Old entries will be unchanged, new entries will be added and existing entries will de updated.
"""

def migrate(cr, version):
    if not version:
        return

    query = (
        "UPDATE ir_model_data"
        " SET noupdate = FALSE "
        " FROM res_bank"
        " WHERE module = 'l10n_nl_bank'"
        "    AND model = 'res.bank'"
        "    AND res_bank.active = TRUE")
    cr.execute(query)
```

StefanRijnhart:
Ah but then we have a misunderstanding about the objective. If the objective is to remove obsolete entries from the datafile, then we should keep noupdate="1" and not have any migration script.

bosd:
Sorry, forgot to add in my previous post that in the datafile I had set the noupdate=1.

1. The datafile should contain only active records. Obsolete records are removed. So it's clean for new installs.
2. On existing installs, all records should be kept in the db. (even when no longer in updated data file)
3. On existing installs new data should be added. Cosmetic changes from the datafile should be populated. (example "knap" changed to "Knab")

StefanRijnhart:
Then leave the noupdate="1" intact, remove obsolete banks from the data file and create a (post) migration for any changes you think are important (your 'knap' example might, but I don't think much of any of the others from this PR).
Note that you are changing the XMLID for KNAP, which you also need to encode in a premigration script.

StefanRijnhart:
Or use openupgradelib to force load the data file. @astirpe is right that it may not be available everywhere, but the reason we made it into a separate lib in the first place was so that it could be used in cases like this. You could attempt an import of openupgradelib in your migration script and proceed to force load the data if available, and in case of an ImportError log an warning saying that the setup is stuck with the old bank data because the lib is missing.

bosd:
I'm finished with the commits.
Don't know why codecov/project is failling and how to fix.
Otherwise LGTM ready for review and merge.

l10n_nl_bank/data/res_bank_data.xml
```diff
@@ -1,219 +1,378 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo noupdate="1">
     <record id="ABNA" model="res.bank">
-        <field name="name">ABN AMRO BANK N.V.</field>
+        <field name="name">ABN AMRO Bank N.V.</field>
         <field name="bic">ABNANL2A</field>
+        <field name="country">165</field>
```
StefanRijnhart: `<field name="country" ref="base.nl"/>` here and below

l10n_nl_bank/migrations/13.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,24 @@
...
+import logging
+
+_logger = logging.getLogger(__name__)
+try:
+    from openupgradelib import openupgrade
+except ImportError:
+    _logger.warning("OpenUpgradeLib is not found, can't update dutch bank data")
+
+
+@openupgrade.migrate(use_env=True)
```
StefanRijnhart: Because openupgrade might not be available. you are not allowed to refer to its decorator in this way. No problem, because the method only uses the cursor which is what you get without the decorator instead of the env.

l10n_nl_bank/__manifest__.py
```diff
@@ -1,14 +1,15 @@
...
     "summary": "Import all Dutch banks with BIC code",
-    "version": "13.0.1.0.0",
-    "category": "Localisation",
+    "version": "13.0.1.1.0",
+    "category": "Localisation/Europe",
     "license": "AGPL-3",
     "author": "Odoo Experts, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/l10n-netherlands",
-    "depends": ["base"],
+    "depends": ["base", "account"],
```
StefanRijnhart: What is it that requires account instead of just base in this change?

bosd: The addition of the account dependency is not specifically related to these changes.
IMO, account should've always been an dependency. Will test further, if not including this dependency returns an error when not installed. Or the module just is useless, because the account fields are unavailable.

l10n_nl_bank/migrations/13.0.1.1.0/post-migration.py
bosd: What is the proper way to solve this?

l10n_nl_bank/__manifest__.py
bosd: Test Results:
Module can be installed without errors when "account" is unavailable.
From functional point of view it is pretty useless to install this module without that one.

Reason for adding this was to be inline with similar modules on other localisation languages. Will remove to give users more freedom for use cases.

bosd:
:raised_hands: Only one item left to fix

l10n_nl_bank/migrations/13.0.1.1.0/post-migration.py
StefanRijnhart: Remove the decorator and change signature to receive the cr that Odoo passes by default (https://github.com/OCA/l10n-netherlands/blob/11.0/l10n_nl_tax_statement/migrations/11.0.2.0.0/post-migrate.py#L8)

bosd: Can't get that to work. It keeps throwing server errors :(

```
def migrate(cr, version):

    openupgrade.load_data(env.cr, "l10n_nl_bank", "data/res_bank_data.xml", mode="init")
```

StefanRijnhart: What kind of errors exactly?

StefanRijnhart: Might want to try passing `cr` to load_data instead of `env.cr` in that case.

bosd: Error:
Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 624, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/opt/odoo/custom/src/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 669, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/auto/addons/web/controllers/main.py", line 1331, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/auto/addons/web/controllers/main.py", line 1319, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 387, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 374, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "<decorator-gen-61>", line 2, in button_immediate_upgrade
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_module.py", line 72, in check_and_log
    return method(self, *args, **kwargs)
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_module.py", line 634, in button_immediate_upgrade
    return self._button_immediate_function(type(self).button_upgrade)
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_module.py", line 573, in _button_immediate_function
    modules.registry.Registry.new(self._cr.dbname, update_module=True)
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 86, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 419, in load_modules
    force, status, report, loaded_modules, update_module, models_to_check)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 315, in load_marked_modules
    perform_checks=perform_checks, models_to_check=models_to_check
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 230, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/opt/odoo/custom/src/odoo/odoo/modules/migration.py", line 180, in migrate_module
    migrate(self.cr, installed_version)
  File "/opt/odoo/auto/addons/l10n_nl_bank/migrations/13.0.1.1.0/post-migration.py", line 23, in migrate
    openupgrade.load_data(env.cr, "l10n_nl_bank", "data/res_bank_data.xml", mode="init")
NameError: name 'openupgrade' is not defined

StefanRijnhart: Ah yes of course. Try something like

```
try:
    from openupgradelib import openupgrade
except ImportError:
    openupgrade = None

def migrate(cr, version):
    if openupgrade is None:
        _logger.warning("OpenUpgradeLib is not found, can't update dutch bank data")
        return
    ...
```

bosd: Maybe you noticed I don't understand exactly what we are doing here. :)
But the version without the requested change was working in both cases. While openupgrade lib was available and when it was unavailable.

StefanRijnhart: Thank you for your patience and your effort!

bosd: Great to be part of this community and learn a lot!
Changes are in the latest commit.

StefanRijnhart: Perfect! Can you squash commits?

StefanRijnhart (approved):

l10n_nl_bank/migrations/13.0.1.1.0/post-migration.py
bosd: Done! :+1:

CasVissers-360ERP (approved):
LGTM

bosd:
Migration to V14 is merged. Let's do this one as well. :smile_cat: 
I don't think he (:robot:) is listening to me. But I try nevertheless 

/ocabot merge nobump

bosd:
> Sorry @bosd you are not allowed to merge.

As expected

> If you wish to adopt an addon and become it's [maintainer](https://odoo-community.org/page/maintainer-role), open a pull request to add your GitHub login to the `maintainers` key of its manifest.

At least I've learned something :smile: 
Well, the maintainers tag is missing from the manifest file. 
But I'm not ready to adopt this puppy :cat2: 

I'm done playing with @OCA-git-bot  :robot:  so @StefanRijnhart  He's all yours.

(I hope will be listening to you today :smile: )

StefanRijnhart:
Oh I'm sorry about that @bosd! But if you keep up the good work on this repo you should be welcome as a maintainer in no time :-)

/ocabot merge nobump

PR #303, opened 2021-01-21
[11.0][IMP] l10n_nl_tax_statement: precreate new columns

StefanRijnhart:
The migration of this module, on a large database, can take hours just to store all those empty values.

### DIFF
--- a/l10n_nl_tax_statement/__init__.py
+++ b/l10n_nl_tax_statement/__init__.py
@@ -1,2 +1,3 @@
+from .hooks import pre_init_hook
 from . import models
 from . import wizard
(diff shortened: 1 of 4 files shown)


CasVissers-360ERP (approved):
Agree, should also be applied to l10n_nl_tax_invoice_basis.

astirpe:
I agree.
I'm wondering if it also takes a long time when installing the module 11.0.2.0.0 for the first time on a large db.
@CasVissers-360ERP do you maybe know something about it?

StefanRijnhart:
@astirpe the mechanism is the same whether installing on a new module or upgrading an existing version of the module without these fields (but having a large NL financial setup that does not already have this *essential* module is much less common I think).

StefanRijnhart:
@CasVissers-360ERP I don't see a stored related field in l10n_nl_tax_invoice_basis, at least not in V11.0

astirpe (approved):
The `pre_init_hook` could be also integrated in version 12 of this module

StefanRijnhart:
~~True, it qualifies for forward-porting~~ Still think that installing this module on a large existing database is less likely because I can't imagine starting a Dutch financial administration without this module.

PR #306, opened 2021-02-26
[13.0][FIX] l10n_nl_xaf_auditfile_export: error bankAccNr length limit exceeded

astirpe:
This fixes the following error:

`Element '{http://www.auditfiles.nl/XAF/3.2}bankAccNr': [facet 'maxLength'] The value has a length of '38'; this exceeds the allowed maximum length of '35'.`

This error could occur in case `acc_number` string contains spaces.


yung-wang (approved):
tested, no more error and file has the correct IBAN

StefanRijnhart (approved):

CasVissers-360ERP (approved):
Seen this issue yes. Thnx for the fix!

PR #307, opened 2021-02-27
[14.0][MIG] l10n_nl_bank

bosd:


CasVissers-360ERP (approved):
LGTM, thnx!

StefanRijnhart:
Ideally, the update from https://github.com/OCA/l10n-netherlands/pull/299 should be a separate commit, cherry picked from your own PR branch (and I agree with renaming the migration scripts for 14.0.1.0.0 because not everybody will have updated the module in 13 before migrating to 14). Would that be alright for you?

bosd:
> Ideally, the update from #299 should be a separate commit

True, It was my intention to do that. Clearly I had not followed the right workflow. Instead of rebasing should've used cherrypicking. Thanks for the hint.

Have been busy lately, will correct this. (But if it takes to long, feel free to merge this one)

> )not everybody will have updated the module in 13 before migrating to 14 Would that be alright for you?

Exactly my line of thinking

bosd:
@StefanRijnhart Done fixing the commit history. Ready for merge now! :tada:

StefanRijnhart (approved):
Perfect, thanks!

astirpe (approved):

PR #268, opened 2020-07-31
[13.0][FIX] l10n_nl_bsn: search bsn with dots

astirpe:
Forward port of latest changes made on https://github.com/OCA/l10n-netherlands/pull/232

PR #309, opened 2021-03-01

astirpe:
Fixes error described in https://github.com/OCA/l10n-netherlands/pull/268#issuecomment-787965251


requirements.txt
```diff
@@ -0,0 +1,2 @@
+# for tests
+vcrpy==2.0.1
```
StefanRijnhart: What's the reason you picked an old version? There is now https://pypi.org/project/vcrpy/4.1.1/. If you have to specify a version at all, you could use `vcrpy<5.0.0` to protect against breaking API changes in the next major release.

astirpe: I just copied it from another repo. I thought there was a reason for specifying that particular version.
Changed to `vcrpy<5.0.0`. Thanks!

StefanRijnhart (approved):

NL66278 (approved):
:+1: LGTM

PR #310, opened 2021-03-02
[12.0][IMP] l10n_nl_tax_statement: by default, don't include moves from before the earliest statement

StefanRijnhart:
Improve performance drastically for older databases by not selecting all historic moves when creating the first tax statement, by setting a minimum default cutoff date for unreported moves to the date of the earliest tax statement (or the statement's "from date" if it is the first).

### DIFF
--- a/l10n_nl_tax_statement/models/l10n_nl_vat_statement.py
+++ b/l10n_nl_tax_statement/models/l10n_nl_vat_statement.py
@@ -199,13 +199,29 @@ class VatStatement(models.Model):
             statement.format_btw_total = btw
 
     @api.model
+    def _get_unreported_move_from_date(self, from_date):
+        """ By default, the unreported date is not set before the start date
+        of the earliest report.
+        :param from_date: the from_date of the statement to compute the
+        unreported date for.
+        """
+        earliest = self.search([], order='from_date asc', limit=1).from_date
+        if not earliest:
+            return from_date
+        offset_date = from_date + relativedelta(months=-3, day=1)
+        return min(from_date, max(earliest, offset_date))
+
+    @api.model
     def default_get(self, fields_list):
         defaults = super().default_get(fields_list)
         company = self.env.user.company_id
         fy_dates = company.compute_fiscalyear_dates(datetime.now())
-        defaults.setdefault('from_date', fy_dates['date_from'])
-        defaults.setdefault('to_date', fy_dates['date_to'])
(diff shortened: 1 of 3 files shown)


l10n_nl_tax_statement/models/l10n_nl_vat_statement.py
```diff
@@ -173,7 +173,16 @@ def _get_company_ids_full_list(self):
...
         compute='_compute_unreported_move_ids'
     )
-    unreported_move_from_date = fields.Date()
+
+    @api.model
+    def _default_unreported_move_from_date(self, date=None):
+        """ Prevent legacy move lines from being selected as unreported by
+        defaulting the unreported date to the start date of the earliest
+        statement.
+        """
```
astirpe: Thanks for this PR!
Is this method missing the body?

astirpe (changes requested):

l10n_nl_tax_statement/models/l10n_nl_vat_statement.py
StefanRijnhart: huh yes! Let me try to find it back...

StefanRijnhart: Remnant of an earlier refactoring. Removed.

astirpe (approved):
Looks good!

l10n_nl_tax_statement/models/l10n_nl_vat_statement.py
```diff
@@ -198,14 +198,30 @@ def _compute_amount_format_btw_total(self):
...
+    def _get_unreported_move_from_date(self, from_date):
+        """ By default, the unreported date is not set before the start date
+        of the earliest report.
+        :param from_date: the from_date of the statement to compute the
+        unreported date for.
+        """
+        earliest = self.search([], order='from_date asc', limit=1).from_date
+        if not earliest:
+            return from_date
+        offset_date = from_date + relativedelta(months=-3, day=1)
```
hbrunn, line 211: where do you get the three months from? My own company's vat duties are quarterly too, but that's not a given, right? Should we either derive the period from existing tax statements, or add a config parameter for the length?

hbrunn: ah, the code below does that already, ignore then

hbrunn (approved):

PR #313, opened 2021-03-11
[8.0][FIX] More livelike dummy object

thomaspaulb:


NL66278:
@thomaspaulb would this be enough to fix a problem where the module field_char_transformed is not available? Maybe possible to return a standard char field, that would just miss the transform function? What errors did you encounter?

thomaspaulb:
I have no idea why I felt I needed this, apparently I ran into an error, but I'm not able to reproduce it now.

PR #314, opened 2021-03-23
[12.0][FIX] l10n_nl_tax_statement: error when fiscal year is present

StefanRijnhart:
When fiscal year entries are present, res.company's compute_fiscalyear_dates
method returns date type values instead of datetime values. Date type values
raise an error when being cast to a date type again.

Regression of #310, which coerces the values to the date type for further
processing.


astirpe (approved):
Thanks!

CasVissers-360ERP (approved):
LGTM

NL66278 (approved):
:+1: LGTM

PR #320, opened 2021-05-04
[12.0][IMP] l10n_nl_vat_statement_icp_line: allow GB ICP lines before brexit

StefanRijnhart:
Our use case is to reconstruct the ICP reports for older periods after a migration (to validate the correctness of that migration wrt. taxes).

### DIFF
--- a/l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py
+++ b/l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py
@@ -51,11 +51,19 @@ class VatStatementIcpLine(models.Model):
 
     @api.constrains('country_code')
     def _check_country_code(self):
-        country_codes = self.mapped('country_code')
+        europe_codes = self.env.ref('base.europe').country_ids.mapped('code')
+        if "GB" not in europe_codes:
+            # Allow GB lines from before the effective Brexit date
+            brexit_date = fields.Date.from_string("2021-01-01")
+            lines = self.filtered(
+                lambda line: line.country_code != "GB" or
+                line.statement_id.from_date >= brexit_date)
+        else:
+            lines = self
+        country_codes = lines.mapped('country_code')
         if self.env.ref('base.nl').code in country_codes:
             raise ValidationError(_(
                 'Wrong country code (NL) for ICP report.'))
-        europe_codes = self.env.ref('base.europe').country_ids.mapped('code')
         for code in country_codes:
             if code not in europe_codes:
                 raise ValidationError(_(


l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py
```diff
@@ -51,11 +51,19 @@ def _compute_icp_amount_format(self):
 
     @api.constrains('country_code')
     def _check_country_code(self):
-        country_codes = self.mapped('country_code')
+        europe_codes = self.env.ref('base.europe').country_ids.mapped('code')
+        if "GB" not in europe_codes:
```
astirpe, line 55: I think [this part](https://github.com/OCA/l10n-netherlands/blob/b7babb2219f8a765d06d9a37e367141cffab9245/l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py#L68-L71) will raise an error in case GB is not in europe_codes.

astirpe (changes requested):

l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py
StefanRijnhart: That part raises an error if 'GB' ends up in country_codes, at least if GB is not a member of the European countries group (which it should not be, by now). So what I'm doing is I'm filtering out lines with code 'GB' from before the brexit date. If there are no GB lines from after brexit, 'GB' will not be in country codes and no error is raised.

astirpe: But when you do `or line.statement_id.from_date >= brexit_date` you expect lines with "GB", right?

StefanRijnhart: Yes, GB lines from after brexit_date will be included in `lines` so that the validation will be applied to them.

astirpe: Shouldn't the raise be skipped in that case?

hbrunn (approved):

l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py
StefanRijnhart: I was aiming to skip the raise for GB lines from *before* the brexit date, not after. I should add tests to demonstrate this, shoudn't I?

astirpe: Sorry, my mistake. You check that `line.statement_id.from_date >= brexit_date` so it is ok as it is.

astirpe (approved):

PR #323, opened 2021-06-09
[14.0][FIX] l10n_nl_xaf_auditfile_export: Rounded move line amounts in template to avoid too many decimals error

jeroen-vet:
Rounded move totals and move line amounts in template as these amounts are not rounded in python code so there could still occur too many decimals errors.


CasVissers-360ERP (approved):
We have been using the same fix for some while now.

thomaspaulb (approved):

PR #325, opened 2021-06-10
[14.0][FIX] Only include move lines that have been posted in the SQL queries

jeroen-vet:
The SQL queries to get opening balance totals, opening balance entries, move line count and move line total debit and credit did not filter for only posted lines. This fix corrects that.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -257,6 +257,7 @@ class XafAuditfileExport(models.Model):
             "     account_account_type t "
             "where a.user_type_id = t.id "
             "and l.account_id = a.id "
+            "and l.parent_state = 'posted' "
             "and l.date < %s "
             "and l.company_id=%s "
             "and t.include_initial_balance = true ",
@@ -278,6 +279,7 @@ class XafAuditfileExport(models.Model):
             "where a.user_type_id = t.id "
             "and a.id = l.account_id and l.date < %s "
             "and l.company_id=%s "
+            "and l.parent_state = 'posted' "
             "and t.include_initial_balance = true "
             "group by a.id, a.code",
             (self.date_start, self.company_id.id),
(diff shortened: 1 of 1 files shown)


thomaspaulb:
@gjotten Are there any cons to not including the unposted lines in the export?

gjotten:
@thomaspaulb it's a workflow thing. We for example have got some moves in a perpetual draft state until year's closing: VAT settlement, profit for the fiscal year, those sorts of things. 

If we were to use an external suite to do monthly/quarterly reports (eg. Visionplanner), then it might be nice to not have to post/unpost those moves on each import. On the other hand, places where I'd expect monthly reporting like that should probably be updating their lock dates and doing at least monthly posting of VAT results.

It might be a wash conceptually, but an accountant will always tell you to only include 'final' moves, and from a business pov the ones doing .xaf exports should be organised enough to handle it.

thomaspaulb (approved):
Code review only.

@gjotten if you agree functionally then you can leave your positive review also.

gjotten:
> Code review only.
> 
> @gjotten if you agree functionally then you can leave your positive review also.

@thomaspaulb to be honest I'd rather test this functionally on a known db, which I don't have for 14.0, and with ways and means to read out the .xaf easily and not be dependent on CTRL+F in xed. The runbot instance as it is right now doesn't even include demodata for account.moves :-(

@NL66278 and @lfreeke maybe you guys can add this branch on your 14.0 migration?

CasVissers-360ERP (approved):
I think required from a functional perspective since some/most external reporting tools expect posted moves.

NL66278 (approved):
:+1: LGTM

thomaspaulb:
Anyone seen this [weird error](https://github.com/OCA/l10n-netherlands/runs/5750195817?check_suite_focus=true) before in another repo and how to solve it?

```
  File "/home/runner/.cache/pre-commit/repo2nxuj_lm/py_env-python3/lib/python3.9/site-packages/black/__init__.py", line 6595, in patch_click
    from click import _unicodefun  # type: ignore
ImportError: cannot import name '_unicodefun' from 'click' (/home/runner/.cache/pre-commit/repo2nxuj_lm/py_env-python3/lib/python3.9/site-packages/click/__init__.py)
```

Perhaps to pin the version of `black` ?

thomaspaulb:
Ah, looks to be [this](https://github.com/psf/black/issues/2964):

```
Please use Black 22.3.0 which was just released to fix this issue or pin Click to 8.0.0, 8.0.1, 8.0.2, 8.0.3, or 8.0.4.
```

PR #326, opened 2021-06-11
[14.0][FIX] XAF_auditfile_export: Solve opening balances mismatch  problem

jeroen-vet:
Currently total debit and credit of opening balances do not match. This is because Odoo has set the attribute `include_initial_balance` for account type 'Current Year Earnings' to True. So with the current code the balances of accounts of that type are included. But these balances as calculated by the sum of their actual entries only show what has been booked away to reserves and do not reflect the (undistributed) profits and losses as Odoo keeps a running total of that automatically. So the difference between debit and credit  exactly amounts to the profits and losses distributed or undistributed to up to the start of the period.
You could solve it by setting include_initial_balance to False and make sure the profits losses of all previous periods have been booked to reserves, but the former will probably have unwanted repercussions and the latter is quite inflexible.
This fix will make sure the sum of credit and debit of opening balances match and will create automatically one opening balance for undistributed profits if need be. So even if the previous period profits have not been distributed it will still be correct.


CasVissers-360ERP (approved):
Fixes incorrect balances. Functional review only.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -251,38 +251,98 @@ def get_taxes(self):
...
+            "and l.account_id = a.id "
+            "and l.parent_state = 'posted' "
+            "and l.date < %s "
+            "and l.company_id=%s "
+            "and (t.include_initial_balance = false or t.id=%s)",
+            (self.date_start, self.company_id.id, cye_type_id),
         )
-        for result in self.env.cr.fetchall():
+        row = self.env.cr.fetchall()[0]
+        if round(row[1] or 0.0, 2) != 0:
```
thomaspaulb, line 335: It's common to use Odoo's `float_compare` or `float_is_zero` in such cases

thomaspaulb (approved):
Code review only.

@gjotten Could you say something about the functional validity of this - I have no clue about these unaffected earnings, neither if they should be placed on the highest account. It looks like it could be valid for some companies and not for others, but I don't know.

gjotten:
@thomaspaulb sounds totally logical to me; previous years' 'results' should not lead to an imbalance between credit/debit when looking at opening balances; it'd mean something got counted double :-)

Since CYE is calculated by Odoo as a choice, it's not an accounting 'fact' insamuch as a way this particular ERP handles things. When we carry over the calculated result, it should either be 'parked' somewhere on the balance sheet as being undistributed, like this PR's fix AFAICS, _or_ we do like @jeroen-vet mentioned by distributing into (an) equity account(s).

PR #328, opened 2021-07-01
[12.0][IMP] l10n_nl_tax_statement_icp: optionally split amounts by product type

StefanRijnhart:
When migrating from a legacy version of Odoo that did not have the dedicated service taxes from later version of the Dutch chart of account, reassigning the service taxes on existing entries or even introducing their usage may not be the most straightforward thing to do. But without it, the amounts on the ICP lines will not be split up correctly between products and services. This PR introduces a parameter that restores the behaviour from the old NL ICP module to split up the amounts based on the product type.

### DIFF
--- a/l10n_nl_tax_statement_icp/README.rst
+++ b/l10n_nl_tax_statement_icp/README.rst
@@ -47,6 +47,18 @@ To configure this module, you need to:
 
 #. Follow the configuration steps as described in *l10n_nl_tax_statement* and set the tag *3b omzet* needed for this report.
 
+On the ICP Statement, the amounts will be split up into an amount for products
+and an amount for services. The Dutch chart of accounts in Odoo provides
+a separate set of taxes for services which ensure that the amounts are split
+up properly here as well. If you are migrating from a legacy situation or you
+do not want to apply separate taxes to service products, you can set a *System
+Parameter* with key
+*l10n_nl_tax_statement_icp.icp_amount_by_tag_or_product* to value *product*.
+
+With this configuration in place, the amounts will be split up according to
+the product type, using service by default when no product is mentioned on
+the invoice line.
+
 Usage
 =====
 
--- a/l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement.py
+++ b/l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement.py
@@ -50,7 +50,6 @@ class VatStatement(models.Model):
                 icp_values['partner_id'] = partner_id
                 icp_values['statement_id'] = statement.id
                 newline = IcpLine.create(icp_values)
-                statement.icp_line_ids += newline
                 icp_total = newline.amount_products + newline.amount_services
                 statement.icp_total += icp_total
 
@@ -135,9 +134,16 @@ class VatStatement(models.Model):
             balance = line.company_currency_id.with_context(
                 date=line.date
             ).compute(balance, self.currency_id, round=True)
+        if self.env["ir.config_parameter"].sudo().get_param(
+                "l10n_nl_tax_statement_icp.icp_amount_by_tag_or_product",
+                "tag") == "product":
+            is_service = not line.product_id or (
+                line.product_id.type == "service")
+        else:
+            is_service = self._is_3b_omzet_diensten_line(line)
         amount_products = balance
         amount_services = 0.0
-        if self._is_3b_omzet_diensten_line(line):
+        if is_service:
             amount_products = 0.0
             amount_services = balance
 
--- a/l10n_nl_tax_statement_icp/__manifest__.py
+++ b/l10n_nl_tax_statement_icp/__manifest__.py
@@ -12,6 +12,7 @@
         'l10n_nl_tax_statement',
     ],
     'data': [
+        'data/ir_config_parameter.xml',
         'security/ir.model.access.csv',
         'views/l10n_nl_vat_statement_view.xml',
         'views/report_tax_statement.xml',
--- a/l10n_nl_tax_statement_icp/data/ir_config_parameter.xml
+++ b/l10n_nl_tax_statement_icp/data/ir_config_parameter.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo noupdate="1">
+    <record model="ir.config_parameter" id="parameter_icp_amount">
+        <field name="key">l10n_nl_tax_statement_icp.icp_amount_by_tag_or_product</field>
+        <field name="value">tag</field>
+    </record>
+</odoo>
--- a/l10n_nl_tax_statement_icp/readme/CONFIGURE.rst
+++ b/l10n_nl_tax_statement_icp/readme/CONFIGURE.rst
@@ -1,3 +1,15 @@
 To configure this module, you need to:
 
 #. Follow the configuration steps as described in *l10n_nl_tax_statement* and set the tag *3b omzet* needed for this report.
+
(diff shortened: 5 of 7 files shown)


l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement.py
```diff
@@ -50,7 +50,6 @@ def _compute_icp_lines(self):
                 icp_values['partner_id'] = partner_id
                 icp_values['statement_id'] = statement.id
                 newline = IcpLine.create(icp_values)
-                statement.icp_line_ids += newline
```
StefanRijnhart, line 53: The lines are already created with a `statement_id`. Removing this line fixes a mysterious `ValueError: Expected singleton: l10n.nl.vat.statement.icp.line(15, 15)` in my test.

hbrunn (approved):

StefanRijnhart:
I see we are hit by https://github.com/odoo/odoo/commit/512abb7a4691b0a961800a341a12dca935fbd29d#r52909262.
Fix should be merged into Odoo 12.0 soon.

l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement.py
```diff
@@ -135,9 +134,16 @@ def _prepare_icp_line_from_move_line(self, line):
             balance = line.company_currency_id.with_context(
                 date=line.date
             ).compute(balance, self.currency_id, round=True)
+        if self.env["ir.config_parameter"].get_param(
```
astirpe: I think that a `sudo()` is needed here

astirpe (changes requested):

astirpe:
@yung-wang what do you think of this option? It allows the compatibility of the ICP module in case the coa does not provide separate taxes for products and services.

yung-wang:
@astirpe For me it sounds more logical to adept to the new logic of service taxes instead making extra option for the product type. This sounds more like a migrating issue and having the customer to adapt the new flow. Just like when you upgrade Odoo core from V11 to V14. You adept the flow to the V14.

l10n_nl_tax_statement_icp/README.rst
```diff
@@ -47,6 +47,18 @@ To configure this module, you need to:
 
 #. Follow the configuration steps as described in *l10n_nl_tax_statement* and set the tag *3b omzet* needed for this report.
 
+On the ICP Statement, the amounts will be split up into an amount for products
+and an amount for services. The Dutch chart of accounts in Odoo provides
+a separate set of taxes for services which ensure that the amounts are split
+up properly here as well. If you are migrating from a legacy situation or you
+do not want to apply separate taxes to service products, you can set a *System
```
astirpe, line 54: @yung-wang you are right, that's something also explained in the readme:

> If you are migrating from a legacy situation or you do not want to apply separate taxes to service products, you can set ...

What do you propose? Do you think this explanation in the readme is not enough?

l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement.py
StefanRijnhart: Thank you for catching this one! Fixed.

StefanRijnhart:
It is not only a practical problem. I also have doubts about assigning a separate tax to service products. It feels as if we are keeping a double administration of what is a service and what is not. The product type should be leading as far as I am concerned.

l10n_nl_tax_statement_icp/README.rst
yung-wang: @astirpe The readme is good. Maybe in the future the migrated customer can adapt to the new flow and stick closer to how standards works. So for now, i approve this.

astirpe (approved):
My approval is for the technical part. For the functional part I rely on the approval of @yung-wang

yung-wang:
> It is not only a practical problem. I also have doubts about assigning a separate tax to service products. It feels as if we are keeping a double administration of what is a service and what is not. The product type should be leading as far as I am concerned.

I agree it would be easier to look only at the product type. But what do you if you don't use products in invoice lines? You still need 2 types of taxes right?

StefanRijnhart:
@yung-wang Am I guessing correctly that you were involved with introducing this tax layout?

```
commit 2eecf81a0e1c5b0dac224125abe9d1c8f0164ff9
Author: Quentin De Paoli <qdp@odoo.com>
Date:   Thu Sep 22 10:50:34 2016 +0200

    [IMP] l10n_nl: New Dutch localization module. Courtesy of Onestein. PR# 12351
```

I would very much like to know what the benefit is of splitting up the taxes across product type. I could not find any specific mentioning of services on https://belastingdienst.nl that warrants this design.

astirpe:
@StefanRijnhart the splitting up of the taxes was proposed as best practice, according to dutch accountants and auditors.

yung-wang:
@StefanRijnhart I was not involved during that time. But if you look at an example ICP opgaaf, you can see there are 2 columns. 1 for services and 1 for products.

So I think 1 of the reason to split them up is:
If you dont use products in invoice lines, you need to have something to define what a service or product is. So logically to me is to use the tax code.
If you use products in invoices lines, only looking at the type would be enough.

https://www.belastingdienst.nl/wps/wcm/connect/bldcontentnl/belastingdienst/zakelijk/btw/zakendoen_met_het_buitenland/goederen_en_diensten_naar_andere_eu_landen/opgaaf_icp/
> Vergelijk de opgaaf ICP met uw btw-aangifte
De leveringen en diensten die u op de opgaaf ICP vermeldt, zijn een specificatie van het bedrag dat u invult in rubriek 3b van uw btw-aangifte. Het totaal van de opgaaf ICP moet dus gelijk zijn aan het bedrag dat u over dezelfde periode invult in uw btw-aangifte bij Rubriek 3b 'Leveringen naar/diensten in landen binnen de EU'.

yung-wang:
@StefanRijnhart but guys, for me this PR is fine.

StefanRijnhart:
OK, I'm not convinced that was the right design given that this is the accounting part of an integrated ERP with the setting on the product, at least not without a mechanism to match the tax product type with the product type on the sale/invoice/move line.

But thank you for your clarifications and reviews!

yung-wang:
> OK, I'm not convinced that was the right design given that this is the accounting part of an integrated ERP with the setting on the product, at least not without a mechanism to match the tax product type with the product type on the sale/invoice/move line.
> 
> But thank you for your clarifications and reviews!

Sure, if you have improvements how to do it better, be free to make a PR. Take into account the 2 scenario's, usages with products and without products.

astirpe:
@StefanRijnhart besides of what @yung-wang already said, in 2016 I was also working together with dutch accountants and auditors who confirmed that the splitting up of the taxes was a correct thing to do on a coa of a professional ERP.

NL66278 (approved):
:+1: LGTM.

StefanRijnhart:
> Sure, if you have improvements how to do it better, be free to make a PR. Take into account the 2 scenario's, usages with products and without products.

Maybe the situation of an intra EU sales invoice without a product on a *non-service* invoice line is already very rare and can, or should be avoided. Under that assumption, the service taxes would be redundant, right?

yung-wang:
> Maybe the situation of an intra EU sales invoice without a product on a _non-service_ invoice line is already very rare and can, or should be avoided. Under that assumption, the service taxes would be redundant, right?

Im not sure. But i can imagine there are people using only Acconting without logistics/timesheet. So they can write directly on an invoice line without linking any product/service to it.

PR #319, opened 2021-04-26
[13.0][FIX] l10n_nl_tax_statement: case no match on tags

astirpe:
Fixes https://github.com/OCA/l10n-netherlands/issues/318

@CasVissers-360ERP thanks for this fix!

PR #334, opened 2021-08-30
[test] don't merge

hbrunn:
rebased version of #319

### DIFF
--- a/l10n_nl_bsn/models/res_partner.py
+++ b/l10n_nl_bsn/models/res_partner.py
@@ -67,6 +67,8 @@ class ResPartner(models.Model):
                 and domain[0] == "bsn_number"
                 and isinstance(domain[2], str)
                 and domain[2]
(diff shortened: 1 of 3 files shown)


hbrunn:
I use this to find out why CI fails in v13, locally I can't reproduce it. Any helpful input is appreciated.

pedrobaeza:
Tests are posting the move twice: one in https://github.com/OCA/l10n-netherlands/blob/13.0/l10n_nl_tax_invoice_basis/tests/test_tax_invoice_basis.py#L51 and for example another in https://github.com/OCA/l10n-netherlands/blob/13.0/l10n_nl_tax_invoice_basis/tests/test_tax_invoice_basis.py#L63

This was not a problem at the beginning of 13.0, but then Odoo put some code to restrict this. Remove all the second calls.

hbrunn:
thank you @pedrobaeza, my OCB was pinned indeed. Fixing PR in https://github.com/OCA/l10n-netherlands/pull/335

PR #343, opened 2022-03-23
[12.0] l10n_nl_xaf_auditfile_export: include lines from inactive journals

antonioburic:
This change makes the contents of the auditfile complete as it includes the lines of inactive journals. 
original contribution by @StefanRijnhart

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -287,9 +287,10 @@ class XafAuditfileExport(models.Model):
     @api.multi
     def get_journals(self):
         '''return journals'''
-        return self.env['account.journal'].search([
-            ('company_id', '=', self.company_id.id),
-        ])
+        return self.env['account.journal'].with_context(
+            active_test=False).search([
+                ('company_id', '=', self.company_id.id),
+            ])
 
     @api.multi
     def get_moves(self, journal):
--- a/l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
@@ -2,12 +2,38 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import base64
+from lxml import etree
 from io import BytesIO
 import os
 from zipfile import ZipFile
 
 from odoo.tests.common import TransactionCase
 from odoo.tools import mute_logger
+
+
+def get_transaction_line_count_from_xml(auditfile):
+    ''' Helper XML method to parse and return the transaction line count '''
+    line_count = 0
+    with ZipFile(BytesIO(base64.b64decode(auditfile)), 'r') as z:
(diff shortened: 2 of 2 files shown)


l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -287,9 +287,10 @@ def get_move_line_total_credit(self):
     @api.multi
     def get_journals(self):
         '''return journals'''
-        return self.env['account.journal'].search([
-            ('company_id', '=', self.company_id.id),
-        ])
+        return self.env['account.journal'].with_context(
+            active_test=False).search([
```
astirpe, line 291: Should this case be covered by a specific test?

antonioburic: @astirpe I was thinking about it, but was unsure if this was too simple for the exact scenario - what do you think, is it a good case to test if the moves of inactive journals are included in the result?

astirpe: Let's see what the authors of this module say...

antonioburic: Added a small test checking the line count after archiving all journals, thanks for the hint @astirpe ^^

astirpe: Thank you @antonioburic . Tests failure do not depend on this PR. I will try to fix them.

l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
```diff
@@ -97,3 +97,21 @@ def test_05_export_success(self):
...
+        all_journals = record.get_journals()
+
+        all_journals.write({'active': False})
+        self.assertTrue(all(not j.active for j in all_journals))
+
+        record_after = self.env['xaf.auditfile.export'].create({})
+        record_after.button_generate()
+        self.assertTrue(record_after)
+
+        move_count_after = record_after.get_move_line_count()
```
StefanRijnhart: I'm afraid this test is not meaningful for the bug, because `get_move_line_count` uses SQL and does not even touch the journal dimension. Maybe you can parse the document using etree and count the number of transaction nodes and then check that it is equal to the result of `get_move_line_count`?

antonioburic: Thanks for the suggestion, modified it in such a way! 🙏

StefanRijnhart:
Thanks for the improved test! I'm wondering why Travis does not run. But I wanted to ask you to squash commits as is conventional in the OCA, so maybe that helps with Travis as well.

astirpe:
Travis fails because of tests of l10n_nl_kvk not passing anymore. I'm working at them.

StefanRijnhart (approved):

astirpe (approved):

astirpe:
@antonioburic you could try to rebase on https://github.com/OCA/l10n-netherlands/pull/345 to make travis green

pankk (approved):

PR #345, opened 2022-03-28
[12.0][IMP] l10n_nl_kvk: new Zoeken API (Search API no longer available)

astirpe:
As of April 1st 2022, the KvK Search API will be no longer available.
This PR implements the switch to new KvK Zoeken API, as described in https://developers.kvk.nl/support/switch-to-new-apis.
This PR also updates the certificate bundle up to the latest version available.


antonioburic (approved):
great job @astirpe! 
tests are failing due to an issue upstream, so nothing from this repo...

StefanRijnhart (approved):
Thanks!

StefanRijnhart:
Tests should be fixed when this ends up in OCB: https://github.com/odoo/odoo/commit/9511700fc5c7d1bc5e57545fc2fdc1ed98e2aef6

l10n_nl_kvk/models/l10n_nl_kvk_api_handler.py
```diff
@@ -90,24 +90,24 @@ def _kvk_http_header(self):
...
         else:
-            url_query += 'testsearch/companies?kvkNumber={0}'
+            url_query = 'https://api.kvk.nl/test/api/v1/'
 
+        url_query += 'zoeken?kvkNummer={0}'
         url_query = url_query.format(kvk)
         return url_query
 
     @api.model
     def _get_url_query_name_api(self, name):
```
NL66278: This method is nearly a copy of the _get_url_query_kvk_api, so perhaps to refactor out the common code in a new method with the key to search on as a parameter.

NL66278 (changes requested):
:+1: Thanks for adapting this module. Apart from my comment on refactoring I also notice that there are hardly or any docstrings in the python code. Maybe a good idea to add these as well.

l10n_nl_kvk/models/l10n_nl_kvk_api_handler.py
astirpe: @NL66278 I grouped the common code. However I would keep the two methods separated, because they are inherited by module `l10n_nl_openkvk`, and so to avoid to change the method signature

NL66278 (approved):
:+1: Thanks for the improvements!

PR #352, opened 2022-05-06
[MIG][15.0] l10n_nl_bank

tarteo:
@espo-tony


NL66278 (approved):
:+1: LGTM

NL66278:
Question to the author of this module: i suppose some kind of script was used to generate the bank data file from the source data. It might be useful to include that script (if that indeed was used) as well to e able to refresh data in the future (or for future maintenance of this module).

tarteo:
If this is the case then it sounds like good idea :+1: Thanks for reviewing

astirpe (approved):
As far as I know, the author of this module is not active in OCA anymore

espo-tony (approved):

bosd:
If such script exsists i am very interested. In the last update I did it manually.

NL66278 (approved):
:+1: LGTM

PR #360, opened 2022-09-09
[FIX][IMP][10.0] l10n_xaf_auditfile_export: Invalid characters removal

thomaspaulb:
XML has some [invalid or discouraged characters](https://stackoverflow.com/questions/1707890/fast-way-to-filter-illegal-xml-unicode-chars-in-python) that the XML validation step fails on. This MR addresses several things:

1. Fix: in case the generated XAF file contains one of these characters (from an Odoo record), it replaces them with a space
2. Usability improvement: In case the XAF generation fails, besides the error, the traceback is also logged in the Odoo logfile
3. Usability improvement: In case the XML validation fails, the auditfile is still saved on the record, so that the user may download it and inspect the line at which the error occurred, and fix his source data. Now he sees a line number with a problem, but can't inspect the actual file to deduce which record caused it.
4. Speed improvement: the method of parsing the XML in order to validate it, resulted in very slow validation for bigger files (more than 5 minutes). Calling `xsd.assertValid` makes the validation, and thus the whole auditfile generation process, much faster (less than 1 minute). [Docs for this here.](https://lxml.de/validation.html#xmlschema)

### DIFF
--- a/l10n_nl_tax_invoice_basis/tests/test_tax_invoice_basis.py
+++ b/l10n_nl_tax_invoice_basis/tests/test_tax_invoice_basis.py
@@ -28,7 +28,7 @@ class TestTaxInvoiceBasis(TransactionCase):
             'account_id': receivable_account_id,
             'date_invoice': '2017-08-18',
             'date': '2017-07-01',
-            'type': 'in_invoice',
+            'type': 'out_refund',
         })
         invoice_line_account_id = self.env['account.account'].search(
             [('user_type_id', '=', self.env.ref(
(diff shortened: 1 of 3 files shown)


NL66278 (approved):
:+1: LGTM Travis errors have nothing to do with this module.

thomaspaulb:
Even so, I fixed that test in https://github.com/OCA/l10n-netherlands/pull/361 and rebased this on it, making Travis green.

NL66278:
@thomaspaulb Travis is green but codecov not. Actually I do not think you can do much about it, except maybe using `#pragma no cover` on the two exception handling parts causing this.

thomaspaulb:
> @thomaspaulb Travis is green but codecov not. Actually I do not think you can do much about it, except maybe using `#pragma no cover` on the two exception handling parts causing this.

I've managed to cover one one of the exception cases, let's see if it's enough. It wasn't a useless exercise fortunately, I found a bug in the "quicker validation" commit during the investigation.

thomaspaulb:
@NL66278 And a refactor got rid of the last issue :-)

PR #374, opened 2023-01-18
[16.0][MIG] l10n_nl_bsn

astirpe:


hbrunn (approved):

l10n_nl_bsn/models/res_partner.py
```diff
@@ -0,0 +1,92 @@
...
+
+            # check is valid, otherwise display a warning
+            warning = self._warn_bsn_invalid()
+
+            # search for another partner with the same BSN
+            args = [("bsn_number", "=", self.bsn_number), ("name", "!=", self.name)]
+
+            # refine search in case of multicompany setting
+            if self.company_id:
+                args += [("company_id", "=", self.company_id.id)]
```
NL66278, line 36: Should this not also includes a search for records where company is False?

l10n_nl_bsn/views/res_partner.xml
```diff
@@ -0,0 +1,17 @@
...
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="res_partner_bsn_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='function']" position="before">
+                <field
+                    name="bsn_number"
+                    attrs="{'invisible': [('company_type','!=', 'person')]}"
```
NL66278: company_type should according to Odoo not be used in business logic. So replace with:
`('is_company', '=', True)`

NL66278 (changes requested):
:+1: Thanks for migrating this. I requested two small changes for things that should have been corrected (IMHO) in earlier versions, but why carry those forward into the future?

l10n_nl_bsn/models/res_partner.py
astirpe: I assume that, in case `if self.company_id` is True, then a different partner having company == False would not conflict, so in this case not any warning is expected. The warning is expected only in case 2 partners have the same explicit company.

l10n_nl_bsn/views/res_partner.xml
astirpe: Right! :+1:

l10n_nl_bsn/models/res_partner.py
NL66278: Let us say we have two companies A and B. There is a partner 'Jan' with bsn '123' (just to lazy here to type a ful bsn) with no company set, this partner is available for company A and B. When we introduce a partner 'Piet' for company A with also bsn '123', both Jan and Piet are now available in company A with the same bsn number. I would count that as a conflict.

astirpe: That's debatable. My original intention was to warn only in case 2 partners have the same explicit company, which is the case that implies a conflict for sure. Instead, in case a user is mixing on purpose partner with companies and without companies with same BSN, that may be something wanted and not necessarily a conflict. In case someone wants to restrict the check, there are 2 options: 1. extend and customize the onchange method or 2. post a change request. At the moment for this migration I don't feel comfortable to modify the functional behavior of this module that turned 7 years old already.

NL66278: Let us leave it as it then for the moment.

NL66278 (approved):
:+1: LGTM

PR #364, opened 2022-10-11
[15.0][MIG]  l10n nl xaf auditfile export

NL66278:

PR #375, opened 2023-01-19
[14.0][FIX] l10n_nl_xaf_auditfile_export: exclude section and note move lines

### LINKED ISSUE #348: [14.0] XAF Audit file improvements
Hi all,

Have some improvements which I think could be beneficial:

1. Do not included move lines with a display_type:
`<trLine t-foreach="m.line_ids" t-as="l"> ->>> <trLine t-foreach="m.line_ids.filtered(lambda l: not l.display_type)" t-as="l">`
2. Add state checks:
add `"and parent_state = 'posted' ", ` to queries:
get_move_line_total_credit
get_move_line_total_debit
get_move_line_count
get_ob_totals
EDIT: There is a PR for this: https://github.com/OCA/l10n-netherlands/pull/325

@antonioburic your work on the module triggered me with some improvements I had in mind.

Cas

astirpe:
Fixes https://github.com/OCA/l10n-netherlands/issues/348

### DIFF
--- a/l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
@@ -169,3 +169,27 @@ class TestXafAuditfileExport(TransactionCase):
         line_count_after = record_after.get_move_line_count()
         parsed_count_after = get_transaction_line_count_from_xml(record_after.auditfile)
         self.assertTrue(parsed_line_count == parsed_count_after == line_count_after)
+
+    def test_07_do_not_include_section_and_note_move_lines(self):
+        """Do not include Section and Note move lines"""
+        self.env["account.move.line"].create(
+            [
+                {
+                    "name": "Section test",
+                    "display_type": "line_section",
+                    "move_id": self.invoice.id,
+                },
+                {
+                    "name": "Note test",
+                    "display_type": "line_note",
+                    "move_id": self.invoice.id,
+                },
+            ]
+        )
+        record = self.env["xaf.auditfile.export"].create({})
+        record.button_generate()
+        self.assertTrue(record)
+
+        line_count = record.get_move_line_count()
+        parsed_line_count = get_transaction_line_count_from_xml(record.auditfile)
+        self.assertEqual(parsed_line_count, line_count)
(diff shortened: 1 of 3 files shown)


CasVissers-360ERP (approved):
LGTM functional review.

l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
```diff
@@ -42,26 +42,27 @@ def get_transaction_line_count_from_xml(auditfile):
 # (esp. product.template).
 @tagged("-at_install", "post_install")
 class TestXafAuditfileExport(TransactionCase):
```
NL66278: The tests are failing because in setUpClass cls.env is indeed not defined. Either keep the test as is or change TransactionCase into SavepointCase. Preferably keep test as is, as in later Odoo released SavepointCase will be deprecated (actually TransactionCase will work then as SavepointCase now).

NL66278: In 16.0 cls.env would be defined in setUpClass...

NL66278 (changes requested):
Please adapt the test class

l10n_nl_xaf_auditfile_export/tests/test_l10n_nl_xaf_auditfile_export.py
astirpe: Yes I see, the cls.env is defined if I set the SavepointCase.
Making use of the setUpClass() was just an attempt to make tests pass. Seems not working, so I'm reverting it and keep the original setUp().
The tests failure is because the SQL code present in the module seems not returning the expected data. It was always working fine in the past but now not anymore. Do you have any idea of why is that happening?

NL66278:
@astirpe The account.move.line's have apparently not been flushed to the database when get_move_line_count is called. I replaced the SQL select with an ORM search_count in the 15.0 migration. That solves the problem. As retrieving a single count will have a neglegible effect on performance (as date is indexed), I think this is the way to solve this. Tried first with a flush(0 on the model before the SQL selecct, but that had no effect. My 15.0 commit is here:
https://github.com/OCA/l10n-netherlands/pull/364/commits/7c121948c978589a68104c1741acc52a8b5af111

astirpe:
@NL66278 it works now, thank you!

NL66278:
@astirpe Please rebase on current 14.0

NL66278:
@astirpe It is missing https://github.com/OCA/l10n-netherlands/commit/7c121948c978589a68104c1741acc52a8b5af111 now.

astirpe:
@NL66278 done, thanks!

NL66278 (approved):
:+1: LGTM

PR #372, opened 2023-01-18
[16.0][MIG] l10n_nl_tax_statement

astirpe:

PR #380, opened 2023-01-25
[16.0][ADD] l10n_nl_tax_statement_date_range

astirpe:
Depends on https://github.com/OCA/l10n-netherlands/pull/372

In `l10n_nl_tax_statement` V16 I removed the date range dependency, to allow the usage of the module without being forced to install `date_range` and its external dependency `odoo_test_helper`.

However, in case there is still any user who needs to set date range, I re-introduce the functionality with this glue module.


l10n_nl_tax_statement_date_range/views/l10n_nl_vat_statement_view.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_l10n_nl_vat_report_form" model="ir.ui.view">
+        <field name="model">l10n.nl.vat.statement</field>
+        <field
+            name="inherit_id"
+            ref="l10n_nl_tax_statement.view_l10n_nl_vat_report_form"
+        />
+        <field name="arch" type="xml">
```
NL66278, line 9: Maybe set from_date and to_date to readonly if date range selected?

hbrunn (approved):

PR #381, opened 2023-02-17

bosd:
Fix the tests for that module


bosd:
This is not the right way to do it. Apparantly even external dependencies which are only used for tests, needs to be listed as external dependencies.

PR #382, opened 2023-02-17

bosd:
Apparantly even external dependencies which are only used for tests, needs to be listed as external dependencies. Otherwise tests will fail with the automatically generated requirements.txt

### DIFF
--- a/l10n_nl_location_nuts/__manifest__.py
+++ b/l10n_nl_location_nuts/__manifest__.py
@@ -10,6 +10,7 @@
     "author": "Onestein, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "depends": ["base_location_nuts"],
+    "external_dependencies": {"python": ["vcrpy"]},
     "post_init_hook": "post_init_hook",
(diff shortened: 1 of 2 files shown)


bosd:
@thomaspaulb @NL66278 Can you please review

thomaspaulb (approved):

NL66278 (approved):
:+1: LGTM

pedrobaeza:
The correct solution is to put `vcrpy` on `test-requirements.txt` file, not this.

bosd:
> The correct solution is to put `vcrpy` on `test-requirements.txt` file, not this.

Ooh, Did'nt knew that.  I've learned something here. :)

@NL66278 Can you revert this? and fix it according to pedro's suggestion?

thomaspaulb:
@bosd Maybe you can prepare a new PR which reverts this and does it in the correct way, then I'll review and @NL66278 can merge again.

NL66278:
@bosd or just a new commit moving it to test-requirements.txt

PR #387, opened 2023-03-17
[14.0] xaf_auditfile_export: improved selection of undivided profit account

NL66278:
Present code always selects the account with the maximum code, even when a specific account has been specified to book the undivided profit on.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -353,7 +353,7 @@ class XafAuditfileExport(models.Model):
             "and l.company_id = %s "
             "and l.parent_state = 'posted' "
             "and t.include_initial_balance = true "
-            "and t.id != %s"
+            "and t.id != %s "
             "group by a.id, a.code",
             (self.date_start, self.company_id.id, cye_type_id),
         )
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -1,4 +1,4 @@
-# Copyright 2015 Therp BV <https://therp.nl>
+# Copyright 2015-2023 Therp BV <https://therp.nl>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import base64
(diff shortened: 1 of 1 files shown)


NL66278:
@thomaspaulb The changes might seem large for a small (but important) effect. However, I had to separate the getting of the amount of undivided profit from the account to book it on. Then it seemed silly to keep the same logic in two places, violating the DRY principle, so I refactored this in the _get_undistributed_profits method. I also put the statements into a global variable the more easy to copy them between psql and the python code to test whether they did what they where supposed to do.

thomaspaulb (approved):
The actual fix is small compared to the overall refactor, but the explanations for the refactors make sense, so: LGTM.

One request, for subsequent reviewers: can you state the description, with example, of

1. What was the expected behaviour
2. What was the actual behaviour, and why was this confusing/wrong
3. How does this fix it

You already did so summarily, but I think an actual example might make the quarter fall more quickly in the mind of a reviewer and then they would be quicker to OK this important fix.

thomaspaulb:
(of course if you have the time, a unit test that actually runs through an example would be even better)

NL66278:
@thomaspaulb The expected result was to see the undivided profit/loss on an account dedicated to that purpose. Instead it was shown on that account that had:
- a transaction before the start date of the auditfile export;
- does not transfer the initial balance;
- had the maximum account code of the accounts that satisfy the two preceding criteria.

Now in the Netherlands and maybe other countries we often give the desired account the maximum code, like '999900'. This is not possible in countries like France where the code, name and purpose of accounts is proscribed by law:
https://fr.wikipedia.org/wiki/Plan_comptable_g%C3%A9n%C3%A9ral_(France)

So the undivided profits would not be booked on '120000 Résultat de l'exercice (bénéfice)' but on '791100 IJSS NETTES SUBROGATION'.  (For a description of what that account is for see: https://payfit.com/fr/fiches-pratiques/comptabilisation-ijss-subrogation/)

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -309,54 +350,68 @@ def get_ob_lines(self):
...
         for result in results:
             yield dict(
                 account_id=result[0],
                 account_code=result[1],
                 balance=round(result[2], 2),
             )
 
+    def _get_undistributed_profits(self):
+        """Get amount of undistributed profits."""
+        cye_type_id = self.env.ref("account.data_unaffected_earnings").id
```
ntsirintanis: Better retrieve the record here, and set the id in the dict like `cye_type_id.id`

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -309,54 +350,68 @@ def get_ob_lines(self):
...
+            ),
+        )
+        row = self.env.cr.fetchone()
+        if not row or not row[0]:
+            return 0.0
+        return round(row[0], 2)
+
+    def _get_undistributed_profit_account(self):
+        """Get the id and the code for the undistributed profit account."""
+        cye_type_id = self.env.ref("account.data_unaffected_earnings").id
```
ntsirintanis: Same as before, it will be more consistent to itself.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -262,42 +318,27 @@ def get_ob_totals(self):
...
+            "and l.company_id = %s "
             "and t.include_initial_balance = true "
-            "and t.id!=%s",
+            "and t.id != %s",
             (self.date_start, self.company_id.id, cye_type_id),
         )
-        row = self.env.cr.fetchall()[0]
+        row = self.env.cr.fetchone()
+        credit, debit, account_count = (
+            (row[0] or 0.0, row[1] or 0.0, row[2] or 0) if row else (0.0, 0.0, 0)
```
ntsirintanis, line 328: Naive question, but how do we know that `row` contains at least 3 elements?

ntsirintanis (approved):
Just a minor comment and a question

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
NL66278: Because there are three columns, and we have an SQL statement that always returns exactly one row.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
NL66278: I changed this one, as well as the previous.

PR #388, opened 2023-03-22
[14.0] l10n nl xaf audit file export reliability

NL66278:
Add more code to prevent server crashes and other problems due to auditfile export

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -7,6 +7,7 @@ import logging
 import os
 import shutil
 import time
+import traceback
 import zipfile
 from datetime import datetime, timedelta
 from io import BytesIO
@@ -202,20 +203,18 @@ class XafAuditfileExport(models.Model):
             del xml
 
             # Validate the generated XML
-            xsd = etree.XMLParser(
-                schema=etree.XMLSchema(
-                    etree.parse(
-                        open(
-                            modules.get_resource_path(
-                                "l10n_nl_xaf_auditfile_export",
-                                "data",
-                                "XmlAuditfileFinancieel3.2.xsd",
-                            )
+            xsd = etree.XMLSchema(
+                etree.parse(
+                    open(
+                        modules.get_resource_path(
+                            "l10n_nl_xaf_auditfile_export",
+                            "data",
+                            "XmlAuditfileFinancieel3.2.xsd",
                         )
                     )
                 )
             )
-            etree.parse(auditfile, parser=xsd)
+            xsd.assertValid(etree.parse(auditfile))
             del xsd
 
             # Store in compressed format on the auditfile record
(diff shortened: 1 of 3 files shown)


l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -211,6 +213,46 @@ def button_generate(self):
...
+        if os.path.exists(lock_file_path):
+            raise exceptions.UserError(
+                _("Generation of auditfile is already in progress.")
+            )
+        lock_info = "User {user} already generating auditfile {auditfile}".format(
+            user=self.env.user.display_name,
+            auditfile=self.name,
+        )
+        with open(lock_file_path, "w+") as tmphandle:
+            tmphandle.write(lock_info)
```
thomaspaulb, line 246: Unfortunately there are Odoo instances around that have multiple servers connecting to the same database, so if you implement a locking mechanism, it's better to have it database-based than file-based.

thomaspaulb: Ok, you're using the filestore folder; clever. Well, let's see what other reviewers say :-)

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -203,6 +203,13 @@ def _get_auditfile_template(self):
 
     def button_generate(self):
         """Generate, store and validate auditfile."""
+        # First check wether file is already there. Should not be possible, because of
+        # the locking, but there is a corner case for single threaded servers, where
+        # this button can be pressed after another sessions has already started the
+        # generation. Then this method will only run after the first job released the
+        # lock and finished processing.
+        if self.auditfile:
+            raise exceptions.UserError(_("Auditfile has already been generated."))
```
thomaspaulb, line 217: Right, but in that single-threaded case, just let it be regenerated a second time.

This case is IMO worse because if you want to regenerate the auditfile you can't?

thomaspaulb:
@NL66278 For clarity to other reviewers, could you edit the description of this issue and add a checkboxed list of PR's that you are forward-porting with this PR?

Also, could you remove the first commit, because that one is already part of another PR.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
NL66278: Already you can't. The generate button is disabled when auditfile is filled. So it is not taking away functionality.

NL66278:
@thomaspaulb When the other PR's are merged, I could rebase this one. Or alternatively when this one is approved we can have all the improvements in one go.

ntsirintanis (approved):
Code review

thomaspaulb (approved):

hbrunn (changes requested):
it's unclear to me which problem this solves.

Then we have a database at hand which is very much specialized in locking things, so the correct way to prevent parallel execution would be something like

`cr.execute(select id from xaf_auditfile_export where id=%s for update', (this.id,))`

instead of some home brewed mechanism.

I like keeping failed auditfiles, but instead of a success flag I'd prefer introducing a state field [draft, done, fail] to be in line with other exports from this repo, like the vat export.

Finally, illegal characters should be fixed with a specialized widget for text output

PR #389, opened 2023-03-30
[FIX] Missing child elements on generalLedger

thomaspaulb:
Fixes `Element '{http://www.auditfiles.nl/XAF/3.2}generalLedger': Missing child element(s). Expected is ( {http://www.auditfiles.nl/XAF/3.2}ledgerAccount`

### DIFF
--- a/l10n_nl_xaf_auditfile_export/views/templates.xml
+++ b/l10n_nl_xaf_auditfile_export/views/templates.xml
@@ -138,8 +138,9 @@
                 <customerSupplierHistory />
             </customerSupplier>
         </customersSuppliers>
-        <generalLedger>
-            <ledgerAccount t-foreach="self.get_accounts()" t-as="a">
+        <t t-set="accounts" t-value="self.get_accounts()" />
+        <generalLedger t-if="accounts">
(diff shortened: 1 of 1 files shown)


astirpe (approved):
LGTM!

ntsirintanis (approved):

l10n_nl_xaf_auditfile_export/views/templates.xml
```diff
@@ -138,8 +138,9 @@
                 <customerSupplierHistory />
             </customerSupplier>
         </customersSuppliers>
-        <generalLedger>
-            <ledgerAccount t-foreach="self.get_accounts()" t-as="a">
+        <t t-set="accounts" t-value="self.get_accounts()" />
```
NL66278, line 141: Should we even allow an export if there are as yet not any accounts for the company?

thomaspaulb: No, but apparently the reason that there were no accounts is that there was another company active in a multi-company setup. So that is still something that needs to be addressed also, maybe with a sudo.

NL66278:
Closed in favor of this one: https://github.com/OCA/l10n-netherlands/pull/391

PR #393, opened 2023-05-15
[14.0][FIX] *+xaf_auditfile_export: use correct company in name

NL66278:
Currently the export record is created with the current company in the environment (self.env.company), but the default company for the user (self.env.user.company_id) is applied to the name generated. This can and will lead to mismatches.

This fix will consistently use self.env.company.


astirpe (approved):

thomaspaulb (approved):

ddejong-therp:
Tested it, and with this fix, the name is now based on the actually selected company.

ddejong-therp (approved):
LGTM

PR #394, opened 2023-06-30
[15.0][IMP] l10n_nl_tax_statement: add XLS export

astirpe:
Adds possibility to export the BTW statement to XLSX

### DIFF
--- a/l10n_nl_tax_statement/tests/test_l10n_nl_vat_statement.py
+++ b/l10n_nl_tax_statement/tests/test_l10n_nl_vat_statement.py
@@ -163,6 +163,18 @@ class TestVatStatement(TransactionCase):
         self.invoice_1 = invoice_form.save()
         self.assertEqual(len(self.invoice_1.line_ids), 5)
 
+    def _check_export_xls(self, statement):
+        """Generate XLS report from action"""
+        report = "l10n_nl_tax_statement.action_report_tax_statement_xls_export"
+        self.report_action = self.env.ref(report)
+        self.assertEqual(self.report_action.report_type, "xlsx")
+        model = self.env["report.%s" % self.report_action["report_name"]].with_context(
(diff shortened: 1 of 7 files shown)


bosd (approved):
LGTM

l10n_nl_tax_statement/tests/test_l10n_nl_vat_statement.py
```diff
@@ -567,3 +567,15 @@ def test_20_multicompany(self):
...
+        """Generate XLS report from action"""
+        report = "l10n_nl_tax_statement.action_report_tax_statement_xls_export"
+        self.report_action = self.env.ref(report)
+        self.assertEqual(self.report_action.report_type, "xlsx")
+        model = self.env["report.%s" % self.report_action["report_name"]].with_context(
+            active_model="l10n.nl.vat.statement"
+        )
+        res = model.create_xlsx_report(self.statement_1.ids, data=None)
+        self.assertTrue(res[0])
+        self.assertEqual(res[1], "xlsx")
```
thomaspaulb: This test now does not result in a lot of coverage, because the report is quite empty.

Is it an idea to modify each existing test to run an XLS export in the end? That way, a lot more cases should be covered.

thomaspaulb (changes requested):

l10n_nl_tax_statement/tests/test_l10n_nl_vat_statement.py
astirpe: I think is a good idea :+1: Thanks!

astirpe: Coverage of `l10n_nl_tax_statement_xlsx.py` increased from 66.15% to 100% :muscle:

thomaspaulb (approved):

PR #396, opened 2023-08-10
[FIX] l10n_nl_xaf_auditfile_export: display_type inconsistency

ddejong-therp:
The SQL queries in l10n_nl_xaf_auditfile_export test whether column `display_type` of `account_move_line` are not in `('line_note', 'line_section')`, like so: `display_type NOT IN ('line_section', 'line_note')`.

However, in SQL, if `display_type` is `NULL`, that condition will always evaluate to `NULL`. This causes some inconsistencies with the numbers found in the XAF report compared to other reports.
A similar thing is not true however for odoo's search domains, so I left those alone.


NL66278 (approved):
:+1: LGTM

astirpe (approved):

Issue #397, opened 2023-08-15
line 1653: b'xmlParseEntityRef: no name'


jerryBlueZebra:
## Module
XAF auditfile export 

## Describe the bug
When i try to generate a audit file it says: line 1653: b'xmlParseEntityRef: no name'

## To Reproduce
Generate a Audit file: Accounting -> Reports -> Audit file export

1. Accounting app
2. Reports
3. Audit file export

**Expected behavior**
Generate a Audit file

gjotten:
To be blunt: this isn't a helpful issue. Which version of Odoo is this for? That's key information we're missing, which should be in the title, which isn't very descriptive anyway. Also, why is this a bug, according to you? OCA is a non-profit collective, not an on-call support team.

Anyway, the problem's with the data being parsed, hence the specific error. And while this is not necessarily a bug with the module itself, you're possibly looking at a lack of proper error management, which I've personally run into with this module as well. Them's the breaks.

When you try to generate an .xaf file with illegal characters you could for example run into this error. Ampersand ('&'), which I suspect is to be found in line 1653 of the temporary .xaf which gets deleted by the module, is one of these chars.

Some massaging of the module code on a testing environment should prevent the temporary .xaf from being deleted. That way you can hunt down the offending record. On the other hand, a query on illegal characters on for example your res.partner table might also solve your problem.

jerryBlueZebra:
I see, thanks for the info.

PR #398, opened 2023-09-20
[16.0][FIX] l10n_nl_xaf_auditfile_export: parse EntityRef error

astirpe:
When an account has an "&" character in its name, the XAF auditfile export fails.

Here are 2 different errors I was able to reproduce:

- `b'xmlParseEntityRef: no name'` (probably the one described in https://github.com/OCA/l10n-netherlands/issues/397)
- `b"EntityRef: expecting ';'"`

Steps to reproduce:

- install any chart of account, in this example we use `l10n_nl`
- create and confirm an invoice
- create an Audifile export and verify that the xaf file is successfully generated
- modify the name of account `110000 Debiteuren` (that for sure is listed in the auditfile), for example name it: "Debiteuren & test"
- create a new Audifile export and verify that an error is raised this time

After some investigations, it seems that `value_to_html()` returns a markup instead of a string. This seems making the `etree.XMLParser()` unhappy in case of any "&" character present. Converting the markup to string solves the issue. Not sure if there's a better way to solve it.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/ir_qweb.py
+++ b/l10n_nl_xaf_auditfile_export/models/ir_qweb.py
@@ -13,7 +13,8 @@ class IrQwebAuditfileStringWidget999(models.AbstractModel):
     @api.model
     def value_to_html(self, value, options):
         value = value[: self._max_length] if value else ""
-        return super().value_to_html(value, options)
+        res = super().value_to_html(value, options)
+        return str(res)  # From markup to string
 
 
 class IrQwebAuditfileStringWidget9(models.AbstractModel):


thomaspaulb:
What do you mean by 'returns a markup' ?

If I look at what I think is the [super](https://github.com/OCA/OCB/blob/16.0/odoo/addons/base/models/ir_qweb_fields.py#L108) function of the one you're modifying (I'm not 100% sure because there are several in `ir_qweb_fields.py`), and I naively run the code in Odoo shell, then I get this:

```
>>> from odoo.tools import pycompat, html_escape
>>> html_escape(pycompat.to_text('a & b'))
'a &amp; b'
>>> str(html_escape(pycompat.to_text('a & b')))
'a &amp; b'
```

I don't see a difference between before and after `str()` - both results are strings.

astirpe:
@thomaspaulb did you try with the steps to reproduce?

thomaspaulb:
@astirpe Not yet, but I believe you that I can reproduce it functionally that way. Before I go into the trouble of replicating the setup and setting a breakpoint to get the answer to my question, I'd ask you first. So, what do you mean by "returns a markup" ?

astirpe:
@thomaspaulb it means that returns a type "Markup" instead of "String".

l10n_nl_xaf_auditfile_export/models/ir_qweb.py
```diff
@@ -13,7 +13,8 @@ class IrQwebAuditfileStringWidget999(models.AbstractModel):
     @api.model
     def value_to_html(self, value, options):
         value = value[: self._max_length] if value else ""
-        return super().value_to_html(value, options)
+        res = super().value_to_html(value, options)
+        return str(res)  # From markup to string
```
thomaspaulb, line 17: I did some tests with the help of https://github.com/OCA/l10n-netherlands/pull/399, and indeed I'm running into the `xmlParseEntityRef: no name` error in cases where a field contains an ampersand.

However your fix only addresses cases where the widget is used. I found that changing `xml.unescape()` to `str(xml)` over [here](https://github.com/OCA/l10n-netherlands/blob/201d1ab5d46236ceb4bb5fb6a415bc9767f989b6/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py#L124) addresses the issue for all fields, regardless of the widget used (mine was [here](https://github.com/OCA/l10n-netherlands/blob/201d1ab5d46236ceb4bb5fb6a415bc9767f989b6/l10n_nl_xaf_auditfile_export/views/templates.xml#L253))

thomaspaulb (changes requested):

l10n_nl_xaf_auditfile_export/models/ir_qweb.py
astirpe: I like your approach, I think is the correct way. Is it ok for you to fix it directly in your PR, so that we can close this one?

thomaspaulb: Let's keep both things separated.

ThijsvOers (approved):
Tested the change and it works now

hbrunn (approved):
I'll merge this as it already fixes part of the problem

/ocabot merge

l10n_nl_xaf_auditfile_export/models/ir_qweb.py
thomaspaulb: Now done in https://github.com/OCA/l10n-netherlands/pull/410

PR #399, opened 2023-09-20
[16.0][FWP] quicker validate, show traceback, keep faulty file, invalid chrs removal

thomaspaulb:
Forward port of #360


NL66278 (approved):
:+1: LGTM and already tested in another version

PR #402, opened 2023-10-19
[16.0][FIX] l10n_nl_tax_statement: hide cash-basis when NL invoice-basis

astirpe:
Hide cash basis when NL invoice basis is True (on the settings).

CC @ThijsvOers


ThijsvOers (approved):
Tested this and ok, cash basis is hidden once NL Tax Invoice basis is set to True

NL66278 (approved):
:+1: LGTM

PR #405, opened 2023-11-13
[14.0][FIX] Do not raise error if protected fields are unchanged.

gfcapalbo:
The standard write would check if protected fields were in dict, if they are, it would check if account.move is not draft and then raise error.
Unfortunately the odoo web client posts to write dict all non-readonly fields in a view, therefore some fields , that will be factually unchanged, are still posted in the dict , therefore raising an error.
When needed, this MR , cleans the dict of those ( inconsequential) fields.

### DIFF
--- a/l10n_nl_tax_statement/models/account_move_line.py
+++ b/l10n_nl_tax_statement/models/account_move_line.py
@@ -33,50 +33,53 @@ class AccountMoveLine(models.Model):
 
     def check_field_is_equal(self, changed_protected_field, values):
         self.ensure_one()
-        if self._fields.get(changed_protected_field) and self._fields[
-            changed_protected_field
-        ].type not in ["many2many", "one2many"]:
-            return self[changed_protected_field] == values[changed_protected_field]
-        # if field is X2M we must translate commands to values.
-        # the only acceptable value is [[6,0,self[changed_protected_field].ids]]
-        # wich is what the web client posts in case there is a editable X2M in form
-        # that is unchanged.
-        return (
-            values[changed_protected_field][0] == 6
-            and values[changed_protected_field][2] == self[changed_protected_field].ids
-        )
+        field = self._fields.get(changed_protected_field)
+        old_value = self[changed_protected_field]
+        new_value = values[changed_protected_field]
+        if field.type in ["many2many", "one2many"]:
+            # if field is X2M , the only acceptable value is
+            # [[6,0,self[changed_protected_field].ids]]
+            # wich is what the web client posts in case there is a editable X2M in form
+            # that is unchanged.
+            # If it is a list of tuple-commands, and the last one is precisely
+            # values[changed_protected_field][2] == self[changed_protected_field].ids
+            # we will not accept any modification to protected fields,
+            # only the standard 6,0,[ids] coming from the default web edit.
+            return (
+                len(values[changed_protected_field]) == 1
+                and values[changed_protected_field][0][0] == 6
+                and values[changed_protected_field][0][2]
+                == self[changed_protected_field].ids
+            )
+        if new_value:
+            return old_value == new_value
+        return bool(old_value) == bool(new_value)
 
     def write(self, values):
         # before doing anything we check the nl_vat_statement:check_state,
-        # if this is not allowed it will raise and no further operations
-        # are necessary.
         if not self._l10n_nl_vat_statement_should_check_write(values):
             return super().write(values)
         # now we add code to check if any of the modified fields present in
         # values are the same as existing field value. this is a known limitation
         # of odoo web client, it passes all non-readonly value fields in a form
         # for writing , even if the value has not been changed.
-        res = True
         protected_fields = self._get_l10n_nl_vat_statement_protected_fields()
         protected_fields_in_values = [x for x in values.keys() if x in protected_fields]
+        invalid_fields = set()
         for this in self:
-            # there are protected fields in dict, we proceed to check if such keys
-            # are in reality leaving the field the same.
-            # we need to recreate this dict every time
-            clean_values = dict(values)
-            for changed_protected_field in protected_fields_in_values:
-                is_equal = this.check_field_is_equal(
-                    changed_protected_field, clean_values
-                )
-                if is_equal:
-                    clean_values.pop(changed_protected_field)
-            # in case of recordset write we need to evaluate write one by one
-            # and re-check that our new clean dict is allowed
-            if this._l10n_nl_vat_statement_should_check_write(clean_values):
-                if this._l10n_nl_vat_statement_should_check_write(clean_values):
-                    this._l10n_nl_vat_statement_check_state()
-                res = res and super(AccountMoveLine, this).write(clean_values)
-        return res
+            for protected_field in protected_fields_in_values:
+                is_equal = this.check_field_is_equal(protected_field, values)
+                if not is_equal:
+                    # if the field is invalid in even one
+                    # of the records it cannot be popped
+                    invalid_fields.add(protected_field)
+        for protected_field in protected_fields_in_values:
+            if protected_field not in invalid_fields:
+                values.pop(protected_field)
+        for this in self:
+            if this._l10n_nl_vat_statement_should_check_write(values):
+                this._l10n_nl_vat_statement_check_state()
+        return super(AccountMoveLine, self).write(values)
 
     @api.model
     def _l10n_nl_vat_statement_should_check_write(self, values):
--- a/l10n_nl_tax_statement/models/account_move_line.py
+++ b/l10n_nl_tax_statement/models/account_move_line.py
@@ -46,10 +46,9 @@ class AccountMoveLine(models.Model):
             # we will not accept any modification to protected fields,
             # only the standard 6,0,[ids] coming from the default web edit.
             return (
-                len(values[changed_protected_field]) == 1
-                and values[changed_protected_field][0][0] == 6
-                and values[changed_protected_field][0][2]
-                == self[changed_protected_field].ids
+                len(new_value) == 1
+                and new_value[0][0] == 6
+                and new_value[0][2] == old_value.ids
             )
         if new_value:
             return old_value == new_value
@@ -77,8 +76,7 @@ class AccountMoveLine(models.Model):
             if protected_field not in invalid_fields:
                 values.pop(protected_field)
         for this in self:
-            if this._l10n_nl_vat_statement_should_check_write(values):
-                this._l10n_nl_vat_statement_check_state()
+            this._l10n_nl_vat_statement_check_state()
         return super(AccountMoveLine, self).write(values)
 
     @api.model
--- a/l10n_nl_tax_statement/models/account_move_line.py
+++ b/l10n_nl_tax_statement/models/account_move_line.py
@@ -31,10 +31,52 @@ class AccountMoveLine(models.Model):
             "tax_tag_ids",
         ]
 
+    def check_field_is_equal(self, changed_protected_field, values):
+        self.ensure_one()
+        if self._fields.get(changed_protected_field) and self._fields[
+            changed_protected_field
+        ].type not in ["many2many", "one2many"]:
+            return self[changed_protected_field] == values[changed_protected_field]
+        # if field is X2M we must translate commands to values.
+        # the only acceptable value is [[6,0,self[changed_protected_field].ids]]
+        # wich is what the web client posts in case there is a editable X2M in form
+        # that is unchanged.
+        return (
+            values[changed_protected_field][0] == 6
+            and values[changed_protected_field][2] == self[changed_protected_field].ids
+        )
+
     def write(self, values):
-        if self._l10n_nl_vat_statement_should_check_write(values):
-            self._l10n_nl_vat_statement_check_state()
-        return super().write(values)
+        # before doing anything we check the nl_vat_statement:check_state,
+        # if this is not allowed it will raise and no further operations
+        # are necessary.
+        if not self._l10n_nl_vat_statement_should_check_write(values):
+            return super().write(values)
+        # now we add code to check if any of the modified fields present in
+        # values are the same as existing field value. this is a known limitation
+        # of odoo web client, it passes all non-readonly value fields in a form
+        # for writing , even if the value has not been changed.
+        res = True
+        protected_fields = self._get_l10n_nl_vat_statement_protected_fields()
+        protected_fields_in_values = [x for x in values.keys() if x in protected_fields]
+        for this in self:
+            # there are protected fields in dict, we proceed to check if such keys
+            # are in reality leaving the field the same.
+            # we need to recreate this dict every time
+            clean_values = dict(values)
+            for changed_protected_field in protected_fields_in_values:
+                is_equal = this.check_field_is_equal(
+                    changed_protected_field, clean_values
+                )
+                if is_equal:
+                    clean_values.pop(changed_protected_field)
+            # in case of recordset write we need to evaluate write one by one
+            # and re-check that our new clean dict is allowed
+            if this._l10n_nl_vat_statement_should_check_write(clean_values):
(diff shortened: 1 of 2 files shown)


gfcapalbo:
@thomaspaulb there will be need to change the tests because our mod does not influence those usecases
coverage may be reduced, waiting for CI results.

gfcapalbo:
@thomaspaulb There was a test failing! but not the one we thought. All other write tests were actually modifying the value of the field, therefore our modification did not influence those tests. but one specific date change test expected a user error while writing the same date as  the one on the account.move.line.  
obviously our mod allowed that. re-wrote the test.  Now increasing coverage to cover our new code branches.

gfcapalbo:
coverage was increased to keep percentage equal.  1 test was modified,  there was a noop (same value) write of a protected field that *should* had raised errors , it now should pass because the value is popped out of the write.  I also added an extra write to test that the same field when writing a non-noop value would raise error.

thomaspaulb (approved):

l10n_nl_tax_statement/models/account_move_line.py
```diff
@@ -31,10 +31,52 @@ def _get_l10n_nl_vat_statement_protected_fields(self):
...
+            clean_values = dict(values)
+            for changed_protected_field in protected_fields_in_values:
+                is_equal = this.check_field_is_equal(
+                    changed_protected_field, clean_values
+                )
+                if is_equal:
+                    clean_values.pop(changed_protected_field)
+            # in case of recordset write we need to evaluate write one by one
+            # and re-check that our new clean dict is allowed
+            if this._l10n_nl_vat_statement_should_check_write(clean_values):
```
NL66278: Why this nested if with the same condition?

l10n_nl_tax_statement/models/account_move_line.py
```diff
@@ -31,10 +31,52 @@ def _get_l10n_nl_vat_statement_protected_fields(self):
             "tax_tag_ids",
         ]
 
+    def check_field_is_equal(self, changed_protected_field, values):
+        self.ensure_one()
+        if self._fields.get(changed_protected_field) and self._fields[
```
NL66278: What happens if a non x2m field in the database (existing value) has a falsey value (0.0 or empty string)? From the code I think it will fail the first condition of the if, and therefore proceed to the part where it checks the x2m field. Or am I mistaken in what happens here?

NL66278: I think the code can be more clear if you first have a simple condition for the type of fields and then handle regular fields and x2m fields separately.

l10n_nl_tax_statement/models/account_move_line.py
```diff
@@ -31,10 +31,52 @@ def _get_l10n_nl_vat_statement_protected_fields(self):
...
+            values[changed_protected_field][0] == 6
+            and values[changed_protected_field][2] == self[changed_protected_field].ids
+        )
+
     def write(self, values):
-        if self._l10n_nl_vat_statement_should_check_write(values):
-            self._l10n_nl_vat_statement_check_state()
-        return super().write(values)
+        # before doing anything we check the nl_vat_statement:check_state,
+        # if this is not allowed it will raise and no further operations
```
NL66278: The comment is basically describing the opposite of what happens. In this case code should speak for itself, so just delete the comments.

l10n_nl_tax_statement/models/account_move_line.py
```diff
@@ -31,10 +31,52 @@ def _get_l10n_nl_vat_statement_protected_fields(self):
...
+        # for writing , even if the value has not been changed.
+        res = True
+        protected_fields = self._get_l10n_nl_vat_statement_protected_fields()
+        protected_fields_in_values = [x for x in values.keys() if x in protected_fields]
+        for this in self:
+            # there are protected fields in dict, we proceed to check if such keys
+            # are in reality leaving the field the same.
+            # we need to recreate this dict every time
+            clean_values = dict(values)
+            for changed_protected_field in protected_fields_in_values:
```
NL66278: You give the variable a name starting with `changed_`  and then test wether it did change at all. Just name the variable `protected_field`.

NL66278: You give the variable a name starting with `changed_`  and then test whether it did change at all. Just name the variable `protected_field`.

l10n_nl_tax_statement/models/account_move_line.py
```diff
@@ -31,10 +31,52 @@ def _get_l10n_nl_vat_statement_protected_fields(self):
...
+                    changed_protected_field, clean_values
+                )
+                if is_equal:
+                    clean_values.pop(changed_protected_field)
+            # in case of recordset write we need to evaluate write one by one
+            # and re-check that our new clean dict is allowed
+            if this._l10n_nl_vat_statement_should_check_write(clean_values):
+                if this._l10n_nl_vat_statement_should_check_write(clean_values):
+                    this._l10n_nl_vat_statement_check_state()
+                res = res and super(AccountMoveLine, this).write(clean_values)
```
NL66278: You can have this out of (after) the loop, and still have the batched write: `return super().write(values)` .

Well, while you are at it, rename the argument `values` in the write method to `vals`  so it is the same as in the Odoo code.

Anyway in this way you do not need the res variable.

l10n_nl_tax_statement/models/account_move_line.py
```diff
@@ -31,10 +31,52 @@ def _get_l10n_nl_vat_statement_protected_fields(self):
...
+        self.ensure_one()
+        if self._fields.get(changed_protected_field) and self._fields[
+            changed_protected_field
+        ].type not in ["many2many", "one2many"]:
+            return self[changed_protected_field] == values[changed_protected_field]
+        # if field is X2M we must translate commands to values.
+        # the only acceptable value is [[6,0,self[changed_protected_field].ids]]
+        # wich is what the web client posts in case there is a editable X2M in form
+        # that is unchanged.
+        return (
```
NL66278: Is this tested? I get the impression, but I might be wrong, that I am missing a level, the value is a series of operations. So basically there should be one operation. Maybe nonsens, but I would expect something like:
```
operations = values[changed_protected_field]
if len operations != 1 or len(operations[0]) != 3:
    return False
return operations[0][0] == 6 and operations[0][2] == self[changed_protected_field].ids
```

NL66278 (changes requested):
See other comments

NL66278 (changes requested):
Also add a test for protected x2many field, one with noop change, one with real change. Maybe for the tax_ids field.

l10n_nl_tax_statement/models/account_move_line.py
astirpe: Indeed the condition seems duplicated:

```python
            if this._l10n_nl_vat_statement_should_check_write(clean_values):
                if this._l10n_nl_vat_statement_should_check_write(clean_values):
```

l10n_nl_tax_statement/models/account_move_line.py
gfcapalbo: No, the write needs to be done this way. We are checking for equality in values.  If we write on a recordset and are checking 

I have tried to do the complete write on "self" when possible, but in this specific case it is  not.

  I need to check the equality on every single record, "clean values" will change for each, and each one needs their separate check.

I could do what you say if  I pop out a protected value ONLY once i have verified that all  protected fields in dict are noop writes, that would work, but I would still need to check one-by one _l10n_nl_vat_statement_check_state  wich is not written for recordsets, but records.


I can remove the "res" though.

gfcapalbo: Cannot do that.
THis checks for equality of protected fields record-per-record, i could do as you say by following the logic "if one record has a non noop protected write , make all fail", but i still need to  if 
               this._l10n_nl_vat_statement_should_check_write(clean_values):

for every record "clean dict" wlll be different and needs to be recreated, so i cannot just re-use it.

Example : 

         date is protected field:    I write 12th october 2023 on a recordset(1,2,3):

            on record 1 it is noop, value was already 12thoct 2023 --: ok
            on record 2 date value is DIFFERENT than 23-oct and state not draft ----> not okay.
            on record 2 date value is DIFFERENT than 23-oct and state not draft ----> not okay.
        
    "clean value" dict  would pop out "date" on rec 1, and then rec 2 and 3 would still be popped out.

   We could do it, but instead of raising error on REALLY Wrong records, it would silently fail sometimes and other times raise error depending on recordset  ordering.


We could implement this by doing:

                If the protected field is valid (equal value) in all records proceed, otherwise raise error, always.

Will do that.

NL66278: I am not suggesting to change anything in the checks, those should be done record by record, just to get the write out of the loop. As a failed check in any record will raise an exception, this will not make any functional difference: one failed records will prevent update of all records.

But with the batched write, this will be more efficient.

l10n_nl_tax_statement/models/account_move_line.py
gfcapalbo: @astirpe thank you. cleaning this up.

l10n_nl_tax_statement/models/account_move_line.py
gfcapalbo: > What happens if a non x2m field in the database (existing value) has a falsey value (0.0 or empty string)? From the code I think it will fail the first condition of the if, and therefore proceed to the part where it checks the x2m field. Or am I mistaken in what happens here?

It will not continue, because the first return always works. 
We first compare all non-m2m and non-o2m, and return that result.

the second return activates only if the field is m2m  or o2m.


As for the 0.0 , empty string case , automatic typecasts seemed to work well , except for "None":
```
>>> 0==False
True
>>> 0.0==False
True
>>> ''==False
False
>>> ''==0-0
False
>>> ''==0.0
False
>>> None==False
False
```

I will explicitly convert them in "Bool" to make this more solid:
```
   >>> bool('')==bool(0.0)
True
>>> bool(None)==bool(0.0)
True
>>> bool(None)==bool(False)
True
```

l10n_nl_tax_statement/models/account_move_line.py
gfcapalbo: i meant : " if this is not allowed  , it will proceed to check and eventually raise and not permit further operations...."
cleaned up.

l10n_nl_tax_statement/models/account_move_line.py
gfcapalbo: done

gfcapalbo: done

l10n_nl_tax_statement/models/account_move_line.py
gfcapalbo: the branch is tested (green in CI)

Yes you are correct it should be [(6,0,[1.2,4])]  so it should be 
values[changed_protected_field][0][0] == 6 

Re-debugging because i remember also:
 values[changed_protected_field][2] == self[changed_protected_field].ids
being correct. will get back to you with explanation or correction.

gfcapalbo: Done

gfcapalbo:
@NL66278 @astirpe all comments addressed , wrote and currently running extra tests for the M2M case on tax_ids change.

l10n_nl_tax_statement/models/account_move_line.py
NL66278: I meant it like this (note that by inversing conditions you can get rid of all the `not`'s, and with some helper values stuff will get more readable):
```
field = self._fields.get(changed_protected_field)
old_value = self[changed_protected_field]
current_value = values[changed_protected_field]
if field.type in ["many2many", "one2many"]:
    return(
          len(current_value) == 1
          and current_value[0][0] == 6
          and current_value[0][2] == old_value.ids
    )
if current_value:
    return old_value == current_value
return bool(old_value) == bool(current_value)
```

l10n_nl_tax_statement/models/account_move_line.py
gfcapalbo: done

l10n_nl_tax_statement/models/account_move_line.py
```diff
@@ -31,10 +31,55 @@ def _get_l10n_nl_vat_statement_protected_fields(self):
...
+            # if field is X2M , the only acceptable value is
+            # [[6,0,self[changed_protected_field].ids]]
+            # wich is what the web client posts in case there is a editable X2M in form
+            # that is unchanged.
+            # If it is a list of tuple-commands, and the last one is precisely
+            # values[changed_protected_field][2] == self[changed_protected_field].ids
+            # we will not accept any modification to protected fields,
+            # only the standard 6,0,[ids] coming from the default web edit.
+            return (
+                len(values[changed_protected_field]) == 1
```
NL66278: Use old_value and new_value also in these lines.

l10n_nl_tax_statement/models/account_move_line.py
```diff
@@ -31,10 +31,55 @@ def _get_l10n_nl_vat_statement_protected_fields(self):
...
+            for protected_field in protected_fields_in_values:
+                is_equal = this.check_field_is_equal(protected_field, values)
+                if not is_equal:
+                    # if the field is invalid in even one
+                    # of the records it cannot be popped
+                    invalid_fields.add(protected_field)
+        for protected_field in protected_fields_in_values:
+            if protected_field not in invalid_fields:
+                values.pop(protected_field)
+        for this in self:
```
NL66278: One more optimization I just see is to switch the conditions, as _l10n_nl_vat_statement_should_check_write(values) has an @api.model decorator, it is the same for all records:
```
if self._l10n_nl_vat_statement_should_check_write(values):
    for this in self:
        this._l10n_nl_vat_statement_check_state()
```

NL66278: An ever more drastic performance enhancement is also possible, but that would involve an even more drastic restructuring of the code:
* Take a subset of the records in self, that do have l10n_nl_vat_statement_id set AND are not in draft state
* Check only for those records whether a protected field has really changed, in that case immediately raise an error (no more needs for an invalid_fields set)
* If an error is not raised, just call super() with the original self

gfcapalbo: true , check_write  is @model, because it conly depends on what keys are present in the write dict. We can keep it out of loop.

First proposal:  we can eliminate that condition all together. We already check that :
```
 59         if not self._l10n_nl_vat_statement_should_check_write(values):                  
 60             return super().write(values)                          
``` 

So i eliminated it altogether:

```
 78         for this in self:                                                               
 79             this._l10n_nl_vat_statement_check_state()                                   
 80         return super(AccountMoveLine, self).write(values)      
```

I  implemented the first  proposal, not the more deep one for this reason-
      
the second restructure, while faster, because it works only on a subset of non-draft state records, does move the logic of _l10n_nl_vat_statement_check_state   outside ,  and put it in this write.
I wrote this with the preference of keeping the logic in the check_state  function...
What we would save is not exponential , but linear.  the check_state is really just a comparison operation.

NL66278: @gfcapalbo You cannot eliminate it altogether. The first check is based on all values passed originally, the second check - that you now eliminated - is based only on the valiues that are changed in one or more records. By eliminating the second check, you undo all the work of checking whether values really changed.

gfcapalbo: @NL66278  Fixed:  https://github.com/OCA/l10n-netherlands/pull/405/commits/2ec172d558c77dc9693e84a6ba2ece7cc34fb086

NL66278 (approved):
:+1: I see still more room for streamlining and one functional enhancements, but I do not want better to kill good enough. Specifically if a protected field is changed in a move line that is in draft, or that is not related to a vat statement, but not changed in the move lines that should not be changed, an error will still be raised unnecessarily. Granted that is a rather theoretical situation. Anyhow merging this as is will help a lot.

PR #406, opened 2023-11-14
[16.0][IMP] l10n_nl_tax_statement_icp: format BTW code on reports

astirpe:
The request for this change comes from an accountant.
It seems that other accounting software do not display the `Landcode` as prefix in the `BTW Nummer`.

Here is just one of the examples:

image

So with this PR, we make the ICP reports resembling the other accounting software.

### DIFF
--- a/l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py
+++ b/l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py
@@ -22,6 +22,9 @@ class VatStatementIcpLine(models.Model):
         string="VAT",
         readonly=True,
     )
+    format_vat = fields.Char(
+        compute="_compute_format_vat",
+    )
     country_code = fields.Char(
         readonly=True,
     )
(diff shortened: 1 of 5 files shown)


l10n_nl_tax_statement_icp/models/l10n_nl_vat_statement_icp_line.py
```diff
@@ -32,6 +35,14 @@ class VatStatementIcpLine(models.Model):
     amount_services = fields.Monetary(readonly=True)
     format_amount_services = fields.Char(compute="_compute_icp_amount_format")
 
+    @api.depends("vat", "country_code")
+    def _compute_format_vat(self):
+        for line in self:
+            if line.country_code and line.vat:
+                line.format_vat = line.vat.lstrip(line.country_code)
+            else:
+                line.format_vat = line.format_vat
```
NL66278: Should this not be: `line.format_vat = line.vat` ?

astirpe: I just wanted to return the same value, but maybe you are right.. is better that way

NL66278 (approved):
:+1: LGTM

gjotten (approved):
Looks logical enough to me! @thomaspaulb do you feel like having a perfunctory look as well?

PR #407, opened 2023-11-30
[9.0][IMP] sanitize forbidden chrs from xml

NL66278:
This is a backport of https://github.com/OCA/l10n-netherlands/pull/360


thomaspaulb:
@NL66278 This needs two things:

- In the description, a link to the PR that this is a backport of
- A dependency on a separate PR that [switches the repo from Travis to GH actions](https://odoo-community.org/groups/contributors-15/contributors-253387), in the hope that this will fix the CI for 9.0. It's a couple of manual steps that usually don't take more than 20 mins or so, I've done it a couple of times.

If the CI won't work even after that, we can probably give up on it and just merge this in.

PR #411, opened 2024-03-18
[12.0][ADD] l10n_nl_xaf_auditfile_export: opening balance support

hbrunn:
cherry pick of the opening balance part from https://github.com/OCA/l10n-netherlands/pull/247

Being at it, I also cherry picked the commit from https://github.com/OCA/l10n-netherlands/pull/323 for rounding non-rounded fields.

Fix for l10n_nl_kvk's tests pending.

### DIFF
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -254,33 +254,31 @@ class XafAuditfileExport(models.Model):
     @api.multi
     def get_ob_totals(self):
         """return totals of opening balance"""
-        self.env.cr.execute(
-            "select sum(l.credit), sum(l.debit), count(distinct a.id) "
-            "from account_move_line l, account_account a, "
-            "     account_account_type t "
-            "where a.user_type_id = t.id "
-            "and l.account_id = a.id "
-            "and l.date < %s "
-            "and l.company_id=%s "
-            "and t.include_initial_balance = true ",
-            (self.date_start, self.company_id.id),
+        result = dict(
+            credit=0.0,
+            debit=0.0,
+            count=0
         )
-        row = self.env.cr.fetchall()[0]
-        return dict(
-            credit=round(row[0] or 0.0, 2),
-            debit=round(row[1] or 0.0, 2),
-            count=row[2] or 0,
-        )
+        for line in self.get_ob_lines():
+            balance = line['balance']
+            if balance > 0:
+                result['debit'] += balance
+            else:
+                result['credit'] -= balance
+            result['count'] += 1
+        return result
 
     @api.multi
     def get_ob_lines(self):
         """return opening balance entries"""
         self.env.cr.execute(
-            "select a.id, a.code, sum(l.balance) "
+            # pylint: disable=sql-injection
+            "select a.id, a.code, sum(l.balance) " +
             "from account_move_line l, account_account a, "
-            "     account_account_type t "
+            "     account_move m, account_account_type t "
             "where a.user_type_id = t.id "
             "and a.id = l.account_id and l.date < %s "
+            "and l.move_id = m.id and m.state = 'posted' "
             "and l.company_id=%s "
             "and t.include_initial_balance = true "
             "group by a.id, a.code",
--- a/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
+++ b/l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
@@ -252,6 +252,48 @@ class XafAuditfileExport(models.Model):
         ])
 
     @api.multi
+    def get_ob_totals(self):
+        """return totals of opening balance"""
+        self.env.cr.execute(
+            "select sum(l.credit), sum(l.debit), count(distinct a.id) "
+            "from account_move_line l, account_account a, "
+            "     account_account_type t "
+            "where a.user_type_id = t.id "
+            "and l.account_id = a.id "
+            "and l.date < %s "
+            "and l.company_id=%s "
+            "and t.include_initial_balance = true ",
+            (self.date_start, self.company_id.id),
+        )
+        row = self.env.cr.fetchall()[0]
+        return dict(
+            credit=round(row[0] or 0.0, 2),
+            debit=round(row[1] or 0.0, 2),
+            count=row[2] or 0,
+        )
+
+    @api.multi
+    def get_ob_lines(self):
+        """return opening balance entries"""
+        self.env.cr.execute(
+            "select a.id, a.code, sum(l.balance) "
+            "from account_move_line l, account_account a, "
+            "     account_account_type t "
+            "where a.user_type_id = t.id "
+            "and a.id = l.account_id and l.date < %s "
+            "and l.company_id=%s "
+            "and t.include_initial_balance = true "
+            "group by a.id, a.code",
+            (self.date_start, self.company_id.id),
+        )
+        for result in self.env.cr.fetchall():
+            yield dict(
+                account_id=result[0],
+                account_code=result[1],
+                balance=round(result[2], 2),
+            )
+
+    @api.multi
     def get_move_line_count(self):
         '''return amount of move lines'''
         self.env.cr.execute(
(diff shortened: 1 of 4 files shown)


l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -251,6 +251,48 @@ def get_taxes(self):
             ('company_id', '=', self.company_id.id),
         ])
 
+    @api.multi
+    def get_ob_totals(self):
+        """return totals of opening balance"""
+        self.env.cr.execute(
+            "select sum(l.credit), sum(l.debit), count(distinct a.id) "
```
StefanRijnhart: Sorry if I missed this in any earlier version I might have reviewed, but this seems to greatly overinflate the numbers in my test setup.

If I receive 1000 euros on my bank account before the start date, and I pay 1000 euros from my bank account before the start date, this 1000 euros should not be included in the summary of the opening balance but here it is included on both the debit and the credit side. Therefore, while the opening balance entries are correct in the exported XML, the totals are not (unless I should be reading the specifications to see that I misunderstand the meaning of these numbers).

hbrunn: @thomaspaulb did you dive into this?

thomaspaulb: @StefanRijnhart might well be right about this, the situation I encountered before adding this was that the opening balances were totally missing from the XAF, which is what I sought to correct with minimal understanding of the issue. I haven't had complaints about it from clients, but that doesn't mean there couldn't be a problem here. @gjotten could you perhaps advise from a more accounting-informed perspective?

gjotten: @StefanRijnhart do you mean that we're at least finding the correct accounts we should be posting an opening balance for, but not the correct balance amounts on them?

Intuitively and in a general sense, thinking that +1000 -1000 should even out into a 0.00 total amount (i.e. balance) at the start of the year would be correct. I can also imagine that such a 0.00 balance on an account should not even trigger posting an entry in the .xaf.

I gather that the .xaf specifications also allow us to get away with simply _either_ having a total amount positive or negative (debit or credit) posted as an amount (=balance) on a single account, but not 0/00?

Are we doing balance calculation somewhere? @thomaspaulb I don't think I see it, but my Pythonese is limited.

StefanRijnhart: If you simply group the balances retrieved in `get_ob_lines` by sign (debit/credit) and sum them you will get the right results.

thomaspaulb: To at least translate the Pythonese, or SQL-ese: what the code is doing is:

- taking any account that according to its account type is supposed to be part of the opening balance
- summing debit and credit for all account move lines prior to the date upon which the opening balance is requested, and including this as TotalDebit and TotalCredit
- summing the balance per account for each move line, and including each of these under `obLines`

Looking now at these [mediocre, second-hand docs](https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/section_1024093953.html#subsect_60094306057) I think what I should have done is:

- taking any account that according to its account type is supposed to be part of the opening balance
- summing the balance per account for each move line, and including each of these under `obLines`
- for the above obLines, check out whether the balance comes out as debit or credit; sum all the debit balances and all the credit balances separately, and include the sums as TotalDebit and TotalCredit

This will lead to way less inflated numbers

thomaspaulb: ... what Stefan says is what I mean, but more concise :)

hbrunn: I'll come up with a version of this tonight, thanks everyone

hbrunn: @gjotten @StefanRijnhart @thomaspaulb now I calculate the totals from the same set as the lines, and filter for posted moves too.

I have a history of getting a knot in my brain from the balance sign of different types of accounts, so please be extra critical there

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -252,39 +252,55 @@ def get_taxes(self):
...
-            "and l.company_id=%s "
-            "and t.include_initial_balance = true "
+            # pylint: disable=sql-injection
+            "select a.id, a.code, sum(l.balance) " +
+            self.get_ob_sql_fragment() +
             "group by a.id, a.code",
-            (self.date_start, self.company_id.id),
         )
         for result in self.env.cr.fetchall():
             yield dict(
```
thomaspaulb, line 288: Hm, it might have been simpler just to have used this iteration of dicts as a source for the totals? But the common source SQL works also.

hbrunn:
tests fixed in https://github.com/OCA/l10n-netherlands/pull/412

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
```diff
@@ -252,39 +252,55 @@ def get_taxes(self):
...
+        ).decode('utf8')
+
+    @api.multi
+    def get_ob_totals(self):
+        """return totals of opening balance"""
+        self.env.cr.execute(
+            # pylint: disable=sql-injection
+            "select sum(l.balance), count(distinct a.id) " +
+            self.get_ob_sql_fragment() +
+            "and l.balance < 0"
```
thomaspaulb: Shouldn't this be `having sum(l.balance) < 0` instead?

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
hbrunn: I've been operating under the idea that we need to support *many* lines here, but you're right, the amount of accounts is the upper bound, and that is low

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
hbrunn: good question, I did this to be consistent with https://github.com/OCA/l10n-netherlands/pull/411/files#diff-f022b656919333e9fc10d33f53e72e0de1db9cd6722667987e4367d6f8d9afc8R130 (make it a credit if balance is negative, debit otherwise) - I don't feel qualified to decide if that's correct

hbrunn: oh wait, I answered the wrong question. But am in the process of rewriting this as of the above anyways, so let's stash this for now

thomaspaulb: It's not about the sign, the sign is alright - I'm talking more of the where vs having thing.

If we state the concise specification of what we wanted to achieve here as Stefan's "... group the balances retrieved in get_ob_lines by sign (debit/credit) and sum them ..." then that's not what you're doing here - here you're summing all negative move line balances and all positive move line balances per account. If you use the `having` notation instead, you'd be summing all balances first, and then using the sign of the result to determine if the sum counts as a credit or a debit. The latter is what I thought we agreed we should do.

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
hbrunn: now I do it like you proposed

l10n_nl_xaf_auditfile_export/models/xaf_auditfile_export.py
thomaspaulb: ... and what your last push changes the code to indeed do.

hbrunn: hmm, wouldn't that mean there's always one of totalCredit or totalDebit 0?

My understanding was we want

totalCredit = sum(//openingBalance/obLine/amnt[../amntTp/text()='C'])
totalDebit = sum(//openingBalance/obLine/amnt[../amntTp/text()='D'])

if this xpath makes sense?

thomaspaulb: Well yes, your xpath is what I meant also. So I think we're in agreement now.

thomaspaulb (approved):

StefanRijnhart (approved):
I get the same numbers as on my balance sheet report now. Thanks!

PR #418, opened 2024-07-22
[16.0][MIG] l10n_nl_partner_name: Migration to 16.0

hbrunn:


l10n_nl_partner_name/models/res_partner.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2017-2022 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    """Extend res.partner with extra fields for Dutch names."""
+
+    _inherit = "res.partner"
+    _l10n_nl_partner_name_infixes = ("van", "der", "den", "op", "ter", "de", "v/d")
```
gjotten: @hbrunn just curious since I've never used this module before personally, but if we use non-defined infixes like `d'`, `'t` or `te`, the helper method below is not going to split names up correctly, right?

hbrunn: exactly. Any other ones I should add?

gjotten: [This wiki page](https://nl.wikipedia.org/wiki/Tussenvoegsel#Lijst_van_(min_of_meer)_Nederlandse_tussenvoegsels) contains a curated summary, as well as a a link to a 12-page list provided by the Interior Ministry.

Looks like the latter should be handled in a reference file if push comes to shove :slightly_frowning_face:

hbrunn: I took the easy way out, added your proposals to the default and made the whole thing configurable. Please review the readme part about that for non-nerd compatibility

bosd:
@hbrunn Can you please retrigger runboat creation??

l10n_nl_partner_name/models/res_partner.py
gjotten: Readme explanation LGTM

hbrunn:
done

gjotten (approved):
LGTM functionally, @NL66278 probably worth a technical review as well.

NL66278 (approved):
:+1: LGTM

PR #427, opened 2025-03-12
[18.0][ADD] l10n_nl_kvk_partner

StefanRijnhart:
**KvK number on partner form**

This module will make the KvK number of a partner available as the field
that was used in Odoo 16.0 and earlier. After Odoo 16.0, the field is
replaced by the generic mechanism using two fields: Peppol Endpoint and
Peppol EAS.

This module will use these fields in the backend to display and store
the KvK number when viewed or altered on the partner form. This can help
when backwards compatibility is required, but having the single field
available also provides better usablility than having to enter the
Peppol Endpoint and Peppol EAS separately (at least for instances or
companies in instances that focus mainly on partners with a KvK number).

To use this module, you can enter the KvK number on the partner form. It
will be stored as the Peppol Endpoint for the partner with the matching
Peppol EAS code.

image

Alternatively, if the eInvoicing Endpoint for the partner is already
filled in with the matching Peppol EAS, the KvK number is visible on the
partner form.


CasVissers-360ERP (approved):
Functional review on runboat

hbrunn (approved):

PR #428, opened 2025-03-17
[18.0][MIG] l10n_nl_tax_statement

astirpe:


StefanRijnhart (commented):
Thanks! Code is looking good. I tried testing with the NL chart of accounts, but unfortunately it is not possible to install it on this project's runboat due to the missing dependency on phonenumbers.

image

Do you think it would be acceptable to add this library as an external dependency of this module?

Relatedly, I was surprised that this module does not depend on l10n_nl. Is there really evidence that this module is in use on instances without l10n_nl?

CasVissers-360ERP (approved):
Functional review

CasVissers-360ERP:
@StefanRijnhart I think no need for a depend on l10n_nl, the module can work without the preloaded configuration (a complete custom CoA)

StefanRijnhart (approved):
Code review + local functional test with l10n_nl

PR #431, opened 2025-05-15
[16.0][ADD] l10n_nl_tax_statement_icp_split

hbrunn:
This module allows users to manage ICP statements separately from BTW
statements.

This is useful when the reporting intervals for both differ, ie BTW
quarterly, but ICP monthly. As a side effect, misconfigured partners wrt
ICP won't hold off creating BTW statements.

To use this module, you need to:

1. Go to Accounting/Reporting/NL ICP statement
2. Handle the same way you'd handle a BTW statement
3. Note there will be an extra column "Error" if some partner is
   misconfigured for ICP


Stephanvandenberg (approved):
This latest change makes this PR very workable - tested & approved.

StefanRijnhart (commented):
Excellent and clever code (as usual), but also clever in the sense that I would appreciate a lo more comments or at least a technical section in the README explaining the approach this module takes and how it alters the relationship with the l10n.nl.vat.statement model. Is that something you would mind to add?

hbrunn:
@StefanRijnhart I've added some text in the latest fixup commit - if you agree that's helpful, I'll squash this?

StefanRijnhart:
@hbrunn Looking good, please squash!

hbrunn:
@StefanRijnhart done

StefanRijnhart (approved):
Thanks!

PR #446, opened 2026-03-26
[16.0][UPD] dotfiles

hbrunn:


.pylintrc
NL66278, line 1: The enabled tests where ordered alphabetically. Is there a reason this is changed?

hbrunn: looks like this is a side effect of them being composed from multiple lists: https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pylintrc.jinja

NL66278 (approved):
👍 LGTM

PR #447, opened 2026-04-18
[19.0][IMP] Update files from updated copier

NL66278:

PR #448, opened 2026-04-21
[19.0][MIG] migrate l10n_nl_tax_statement

NL66278:
Created my first tax report for Q1 2026 with it, so works in practice.


hbrunn (commented):
does this need to include #447?

hbrunn:
> does this need to include #447?

probably because of the pylint checks removed there. But I'm hesitant to merge this as it misses caching for pre-commit

NL66278:
@hbrunn The line for the pre-commit cache, that was originally included, caused this error: `2026-04-18T09:29:49.9619806Z WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag`. Then when comparing it with other repo's I did not see it. Removing the pre-commit cache made the error go away.

NL66278:
> > does this need to include #447?
> 
> probably because of the pylint checks removed there. But I'm hesitant to merge this as it misses caching for pre-commit

And indeed, this was because of the pylint checks. Note that #447 fails test, because they do not accept a repository without a module. See the remaining errors there. As Janik helpfully explained to me on the contributors list:
```odoo.conf validation has become stricter with Odoo 19.0.

If none of the modules in a module folder have __main__.py the fodler is skipped.```

hbrunn:
> The directory '/github/home/.cache/pip'...

that looks like a warning, does this really fail the step? I'm very reluctant to merge a deviation from the copier template, because this will cause trouble on the next copier update.

NL66278:
> > The directory '/github/home/.cache/pip'...
> 
> that looks like a warning, does this really fail the step? I'm very reluctant to merge a deviation from the copier template, because this will cause trouble on the next copier update.

It fails on warnings because of this line in the copier-answers: `enable_checklog_odoo: true`. This I took unchanged from the 18.0 branch. I could disable it, or add the specific warnings to the ignore list in `checklog-odoo.cfg`. But would this second option then not be problematic with a new copier update?

hbrunn:
that would indeed be a problem. But I don't see how checklog-odoo has anything to do with warnings in the pre-commit run?

on another note, I just updated crowdfunding to v1.40 and there the config falling out of the template just works: https://github.com/OCA/crowdfunding/actions/runs/24878831168/job/72841882177, so I invite you to try this again

NL66278:
@hbrunn I added back-in the pre-commit line and now it is working without warning. Baffles me why it failed in the first place then.

hbrunn (approved):

StefanRijnhart (approved):
Thanks!

PR #454, opened 2026-08-31
[16.0][MIG] l10n_nl_business_document_import

bosd:
Migration of `l10n_nl_business_document_import` from 14.0 to 16.0.

The module adds partner matching on the Dutch chamber of commerce number (KvK)
and on the OIN to `base_business_document_import` (OCA/edi).

### Changes beyond the plain version bump

- **`partner_coc` is gone in 16.0.** It was dropped from OCA/partner-contact,
  and `partner_identification` ships a `16.0.1.0.0/post-migration.py` that moves
  its value into the native `res.partner.company_registry` field. The CoC (KvK)
  matching therefore uses `company_registry`, and the `partner_coc` dependency
  is removed.
- **`l10n_nl_oin` renamed its field** `nl_oin` -> `l10n_nl_oin` in 16.0. The
  search and the `partner_dict` key follow that rename.
- **`l10n_nl_oin` is no longer a hard dependency.** The OIN lookup is guarded by
  a check on `res.partner._fields`, so the module still works — matching on the
  KvK number only — when `l10n_nl_oin` is not installed. Installing
  `l10n_nl_oin` enables the OIN match without any further action.
- The manifest now depends on `l10n_nl` so that `auto_install: True` does not
  pull the module into non-Dutch databases (previously that was implicit via the
  `l10n_nl_oin` dependency).

### Tests

Rewritten into four cases: KvK match, OIN match, KvK taking precedence over OIN,
and empty values falling through to the standard name matching. The two OIN
cases `skipTest` when `l10n_nl_oin` is not installed. The old reliance on
`base.res_partner_4` demo data is dropped; all records are created by the test.

### Depends on


StefanRijnhart (changes requested):

StefanRijnhart (changes requested):
I don't really understand or agree with the choices in this migration so I would like to encourage you to apply more human intelligence to your submissions.

* if chamber of commerce is now encoded in a standard Odoo field from the `base` module, the integration with base_business_document_import does not belong in l10n-netherlands.
* with the dependency on `l10n_nl_oin` relaxed, what is the use of substituting the dependency with `l10n_nl`? But then again, with the nl-specific CoC field obsoleted, this module could simply be renamed to l10n_nl_oin_business_document with the dependency restored.

Error in PR description: nl_oin was renamed in 15.0, not 16.0.

StefanRijnhart:
Actually, given that account_peppol was backported to Odoo you might want to keep the current module name and map the field names to the peppol endoint schemes like you do in later versions of the module. It would not be unreasonable to add a dependency on account_peppol.

One more basic question about this module: how do `l10n_nl_oin` and/or `l10n_nl_kvk` actually end up in the `partner_dict`?

PR #455, opened 2026-08-31
[17.0][MIG] l10n_nl_business_document_import

bosd:
Migration of `l10n_nl_business_document_import` from 16.0 to 17.0.

Depends on #454 (the 16.0 migration).

### The Dutch KvK and OIN moved to the Peppol fields

As of Odoo 17 the chamber of commerce number (KvK) and the
Organisatie-identificatienummer (OIN) are no longer held in a dedicated field.
They are stored as the generic `peppol_endpoint` of the partner, with
`peppol_eas` stating which of the two the number is. `EAS_MAPPING` in core
`account_edi_ubl_cii` declares `{"NL": {"0106": None, "0190": None}}` — `0106`
is the KvK, `0190` the OIN. This repo's own `l10n_nl_kvk_partner` exists for the
same reason.

The partner matching therefore performs three lookups, tried before the
standard matching on contact, name, website and e-mail:

| `partner_dict` key | matched on |
| --- | --- |
| `company_registry` | `res.partner.company_registry`, as in 16.0 |
| `l10n_nl_kvk` | `peppol_endpoint` with `peppol_eas = 0106` |
| `l10n_nl_oin` | `peppol_endpoint` with `peppol_eas = 0190` |

The EAS code is always part of the search, so a KvK number will not match a
partner that has the same digits registered as an OIN. There is a test for
exactly that.

### On the l10n_nl_oin dependency

The 16.0 version guarded an optional dependency on `l10n_nl_oin`. That module
does not exist in 17.0, and the OIN is now reachable through the Peppol fields,
so the guard is gone and `account_edi_ubl_cii` (core Odoo) is a plain
dependency instead.

### Depends on


StefanRijnhart (changes requested):

PR #456, opened 2026-08-31
[18.0][MIG] l10n_nl_business_document_import

bosd:
Migration of `l10n_nl_business_document_import` from 17.0 to 18.0.

Depends on #455 (the 17.0 migration). No functional change with respect to it.

### Partner matching

The Dutch chamber of commerce number (KvK) and the
Organisatie-identificatienummer (OIN) are held as the generic
`peppol_endpoint` of the partner, with `peppol_eas` stating which of the two
the number is. `EAS_MAPPING` in core `account_edi_ubl_cii` declares
`{"NL": {"0106": None, "0190": None}}`.

Three lookups are tried before the standard matching on contact, name, website
and e-mail:

| `partner_dict` key | matched on |
| --- | --- |
| `company_registry` | `res.partner.company_registry` |
| `l10n_nl_kvk` | `peppol_endpoint` with `peppol_eas = 0106` |
| `l10n_nl_oin` | `peppol_endpoint` with `peppol_eas = 0190` |

The EAS code is always part of the search, so a KvK number will not match a
partner that has the same digits registered as an OIN.

`l10n_nl_kvk_partner` in this repository can be used to display and edit the
KvK number as a single field on the partner form; matching on the Peppol fields
covers partners maintained either way.

### Notes on the port

This module keeps the standard migration version `18.0.1.0.0`.

`_hook_match_partner(partner_dict, chatter_msg, domain, order)` is unchanged in
`base_business_document_import` between the 17.0 and 18.0 series, and
`odoo.osv.expression` is still the API used there, so the override ports
unchanged. What differs in that dependency between the two series is the Odoo 18
translation API (`_()` to `self.env._()`) and a new `user_error_wrap`
signature, neither of which affects this module.

### Depends on


StefanRijnhart (changes requested):

bosd:
@StefanRijnhart Thanks! You're very quick!! I'm still working on it.

